PR #8, opened 2019-02-24
fix: arguments type for compiled function

boopathi:
Fixes type errors in the editor configured with latest typescript which includes types for call, bind and apply - at `func.apply(null, [...])`


boopathi:
👍

ruiaraujo:
:+1:

PR #9, opened 2019-02-24
Removed custom ExecutionResult type.

ruiaraujo:
With the work done on #3, there is no need for the custom type.


ruiaraujo:
:+1:

angelaigreja:
:+1:

PR #14, opened 2019-04-07
Add support for isTypeof on Object type.

ruiaraujo:
It will validate objects before using them if defined. It will not support promises for now.

### DIFF
--- a/src/execution.ts
+++ b/src/execution.ts
@@ -92,6 +92,7 @@ const GLOBAL_EXECUTOR_NAME = "__executor";
 const GLOBAL_ROOT_NAME = "__rootValue";
 const GLOBAL_VARIABLES_NAME = "__variables";
 const GLOBAL_CONTEXT_NAME = "__context";
+const GLOBAL_INSPECT_NAME = "__inspect";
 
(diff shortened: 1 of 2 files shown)


src/execution.ts
```diff
@@ -604,23 +610,41 @@ function compileLeafType(
...
   fieldMap: { [key: string]: FieldNode[] },
   alwaysDefer: boolean
 ): string {
-  let body = `{`;
+  let body = "(";
+  if (typeof type.isTypeOf === "function" && !alwaysDefer) {
+    context.dependencies.set(type.name + "IsTypeOf", type.isTypeOf);
+    body += `!${type.name}IsTypeOf(${originPaths.join(
+      "."
+    )}) ? (${errorDestination}.push(${createErrorObject(
```
boopathi: this could be an if..else. what do you think ... both branches are used as statements rather than expressions.

ruiaraujo: This is inside a giant expression so I can't really change it in this PR.

boopathi:
👍

ruiaraujo:
:+1:

PR #15, opened 2019-04-07
Add support for skipping generating stack traces.

ruiaraujo:
Stack generation is an expensive operation that might be skippable for certain scenarios.

### DIFF
--- a/src/execution.ts
+++ b/src/execution.ts
@@ -55,6 +55,11 @@ export interface CompilerOptions {
     // only safe for use if the output is completely correct.
     disableLeafSerialization: boolean;
 
+    // Disable builtin scalars and enum serialization
+    // which is responsible for coercion,
+    // only safe for use if the output is completely correct.
+    captureStackErrors: boolean;
+
     // Map of serializers to override
     // the key should be the name passed to the Scalar or Enum type
     customSerializers: { [key: string]: (v: any) => any};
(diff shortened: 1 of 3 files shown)


src/execution.ts
```diff
@@ -775,7 +789,7 @@ function unpromisify(
         value
             .then(
                 (res: any) => cb(res, null),
-                (err: Error) => (err != null ? cb(null, err) : cb(null, new Error("")))
+                (err: Error) => (err != null ? cb(null, err) : cb(null, new (CustomGraphQLError as any)("")))
```
topicus: You can do something like:
```
class CustomGraphQLError {
  constructor(message) {
    this.message = message;
    this.name = "CustomGraphQLError";
    return Object.setPrototypeOf(this, Error.prototype);
  }
  toString() {
    return `${this.name}: ${this.message}`;
  }
}
```
then:

```
const e = new CustomGraphQLError("Some error");
console.log(e instanceof Error) // true
```
The performance penalty is minimal.

boopathi: Use `GraphqlJitError`. or `GraphQLError`

src/execution.ts
```diff
@@ -55,6 +55,11 @@ export interface CompilerOptions {
     // only safe for use if the output is completely correct.
     disableLeafSerialization: boolean;
 
+    // Disable builtin scalars and enum serialization
+    // which is responsible for coercion,
+    // only safe for use if the output is completely correct.
+    captureStackErrors: boolean;
```
vigneshshanmugam: nit: comments is unrelated to the flag.

ruiaraujo: not a nit. Good catch :+1:

ruiaraujo:
:+1:

topicus:
:+1:

PR #18, opened 2019-04-14
Fix reused object in variable parsing.

ruiaraujo:
The default value object was created once on compilation and reused on every query.
This was a mistake that is harmless when all values are always specified.

When one of the inputs is sometimes not present, then this bugs show up with the previous value that it was set to.

This PR correctly creates the object inside the instantiated function.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "graphql-jit",
-  "version": "0.1.1",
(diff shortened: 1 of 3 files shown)


ruiaraujo:
:+1:

angelaigreja:
:+1:

PR #22, opened 2019-04-16
feat: provide expanded fields in resolve info for lookaheads

boopathi:
The PR computes a representation of the fields in the Query to a tree form with all the fragments expanded. This is done at the compilation time and this info is useful at runtime to get the fields requested for a node in the graph. 
Shape of the result:

```ts
interface GraphQLJitResolveInfo extends GraphQLResolveInfo {
  fieldExpansion: FieldExpansion;
}
interface FieldExpansion {
  [returnType: string]: TypeExpansion
}
interface TypeExpansion {
  [fieldName: string]: FieldExpansion | { [LeafField]: true }
}
```

This is a helper for quick access to the list of fields selected at the current level. This is useful for transformations into queries to corresponding backends that can accept a list of fields for optimizing the amount of data and other computations.

Example:

```graphql
type Query {
  iBar: IBar
}
interface IBar {
  id: ID!
  title: String
}
type Bar1 implements IBar {
  id: ID!
  title: String
  b1: Int!
}
type Bar2 implements IBar {
  id: ID!
  title: String
  b2: Boolean!
}
```

and the query

```graphql
query {
  iBar {
    id
    title
    ... on Bar1 {
      b1
    }
    ... on Bar2 {
      b2
    }
  }
}
```

the `resolveInfo` at `iBar` would contain the following `fieldExpansion`: 

```
Object {
  "Bar1": Object {
    "b1": true,
    "id": true,
    "title": true,
  },
  "Bar2": Object {
    "b2": true,
    "id": true,
    "title": true,
  },
  "IBar": Object {
    "id": true,
    "title": true,
  },
}
```

### Advantages

This way, at runtime, we do get a list of fields that we can use to query the provider of `iBar` in the above example and it saves us in the following way -

1. response time of Provider
1. amount of data transfer
1. Deserialization time (say, JSON.parse time) - which are cpu bound.

### DIFF
--- a/src/__tests__/resolve-info.test.ts
+++ b/src/__tests__/resolve-info.test.ts
@@ -1,0 +1,181 @@
+import { GraphQLSchema, DocumentNode, parse } from "graphql";
+import { GraphQLJitResolveInfo } from "../resolve-info";
+import { compileQuery, isCompiledQuery } from "../execution";
+import { makeExecutableSchema } from "graphql-tools";
+
+describe("resolve-info", () => {
+  describe("GraphQLJitResolveInfo - simple types", () => {
+    let inf: any;
+    const schema = makeExecutableSchema({
+      typeDefs: `
+        type Query {
+          foo: Foo
+        }
+        type Foo {
+          a: String
+          b: Int
+          c: Boolean!
+        }
+      `,
+      resolvers: {
+        Query: {
+          foo(_: any, _1: any, _2: any, info: any) {
+            inf = info;
+          }
+        }
+      }
+    });
+
+    afterEach(() => {
+      inf = undefined;
+    });
+
+    test("all selection fields of the current resolver", async () => {
+      await executeQuery(schema, parse(`query { foo { a b c } }`));
+      expect(inf.fields).toMatchObject({
+        Foo: expect.arrayContaining(["a", "b", "c"])
+      });
+    });
+
+    test("interface fields", async () => {
+      await executeQuery(
+        schema,
+        parse(
+          `
+          query {
+            foo {
+              ...fooFragment1
+              a
+              b
+              ... on Foo {
+                c
+                ...fooFragment1
+              }
+            }
+          }
+          fragment fooFragment1 on Foo {
+            a
+            b
+          }
+          `
+        )
+      );
+
+      expect(inf.fields).toMatchObject({
+        Foo: expect.arrayContaining(["a", "b", "c"])
+      });
+    });
+  });
+
+  describe("jit resolve info for interfaces", () => {
+    let inf: any;
+    const schema = makeExecutableSchema({
+      typeDefs: `
+        type Query {
+          iBar: IBar
+        }
+        interface IBar {
+          id: ID!
+          title: String
+        }
+        type Bar1 implements IBar {
+          id: ID!
+          title: String
+          b1: Int!
+        }
+        type Bar2 implements IBar {
+          id: ID!
+          title: String
+          b2: Boolean!
+        }
+      `,
+      resolvers: {
+        Query: {
+          iBar(_: any, _1: any, _2: any, info: any) {
+            inf = info;
+          }
+        },
+        IBar: {
+          __resolveType() {
+            return "Bar1";
+          }
+        }
+      }
+    });
+
+    afterEach(() => {
+      inf = undefined;
+    });
+
+    test("compute interface field nodes", async () => {
+      await executeQuery(
+        schema,
+        parse(
+          `
+            query {
+              iBar {
+                id
+                title
+              }
+            }
+          `
+        )
+      );
+
+      expect(inf.fields).toMatchObject({
+        IBar: expect.arrayContaining(["id", "title"])
+      });
+    });
+
+    test("fields per type", async () => {
+      await executeQuery(
+        schema,
+        parse(
+          `
+            query {
+              iBar {
+                id
+                title
+                ... on Bar1 {
+                  b1
+                }
+                ... on Bar2 {
+                  b2
+                }
+              }
+            }
+          `
+        )
+      );
+
+      expect(inf.fields).toMatchObject({
+        IBar: expect.arrayContaining(["id", "title"]),
+        Bar1: expect.arrayContaining(["id", "title", "b1"]),
+        Bar2: expect.arrayContaining(["id", "title", "b2"])
+      });
+    });
+  });
+});
+
+function executeQuery(
+  schema?: GraphQLSchema,
+  document?: DocumentNode,
+  rootValue?: any,
+  contextValue?: any,
+  variableValues?: any,
+  operationName?: string
+) {
+  const prepared: any = compileQuery(
+    schema as any,
+    document as any,
+    operationName || ""
+  );
+  if (!isCompiledQuery(prepared)) {
+    return prepared;
+  }
+  return prepared.query(rootValue, contextValue, variableValues || {});
+}
+
+interface Resolver {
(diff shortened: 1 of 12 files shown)


src/resolve-info.ts
```diff
@@ -0,0 +1,201 @@
...
+  collector: SelectionFieldsCollector
+) {
+  switch (node.kind) {
+    case "Field":
+      collectorSet(collector, returnType, node);
+      break;
+
+    case "InlineFragment":
+      if (node.typeCondition == null) {
+        throw new Error(`TypeCondition is undefined for ${node}`);
```
boopathi: This should not happen. Wonder why the type definitions for this says that it's possibly undefined. Can you have an InlineFragment in GraphQL without a TypeCondition?

boopathi:
```graphql
query {
  foo {
    ... { # this is an InlineFragment without typeCondition
      a b c
    }
  }
}
```

src/__tests__/resolve-info.test.ts
```diff
@@ -0,0 +1,417 @@
...
+              }
+            }
+          `
+        )
+      );
+
+      expect(inf.fields).toMatchObject({
+        Foo: expect.arrayContaining(["foo"]),
+        Bar: expect.arrayContaining(["bar"])
+      });
```
ruiaraujo: Why not a stricter equal?

src/resolve-info.ts
```diff
@@ -0,0 +1,320 @@
...
+  isInterfaceType,
+  isListType,
+  isNonNullType,
+  isObjectType,
+  SelectionNode,
+  SelectionSetNode
+} from "graphql";
+import { ObjectPath } from "./ast";
+
+export interface GraphQLJitResolveInfo extends GraphQLResolveInfo {
```
ruiaraujo: This should be re-exported from the index.

src/resolve-info.ts
```diff
@@ -0,0 +1,320 @@
...
+}
+
+/**
+ * Handle different kinds of selection nodes accordingly.
+ *
+ * For a field, add the field to the TypeExpansion. Create and link
+ * the next FieldExpansion object to be handled recursively
+ *
+ */
+function handleSelection(
```
ruiaraujo: We should file an issue that `@skip` is not supported to be fixed later.

boopathi: https://github.com/ruiaraujo/graphql-jit/issues/17

src/resolve-info.ts
```diff
@@ -0,0 +1,247 @@
...
+  fieldType: GraphQLOutputType;
+  fieldName: string;
+  fieldNodes: FieldNode[];
+}) {
+  const returnType = memoizedResolveEndType(fieldType);
+
+  // Result
+  const fieldExpansion: FieldExpansion = {};
+
+  if (returnType != null) {
```
ruiaraujo: When can this be null?

src/resolve-info.ts
```diff
@@ -0,0 +1,247 @@
...
+    for (const fieldNode of fieldNodes) {
+      deepMerge(
+        fieldExpansion,
+        memoizedExpandFieldNode(schema)(fragments)(fieldNode)(returnType)
```
ruiaraujo: These 4 arguments memoizations do seem to have much value.

I expect `fieldNode` to be different.

src/__tests__/resolve-info.test.ts
```diff
@@ -0,0 +1,803 @@
...
+      `);
+    });
+
+    test("__typename", async () => {
+      const result = await executeQuery(
+        schema,
+        parse(`
+          query {
+            uBaz {
+              __typename
```
ruiaraujo: Why are we excluding `__typename`?

src/__tests__/resolve-info.test.ts
```diff
@@ -0,0 +1,417 @@
+import { GraphQLSchema, DocumentNode, parse } from "graphql";
+import { compileQuery, isCompiledQuery } from "../execution";
+import { makeExecutableSchema } from "graphql-tools";
+
+describe("GraphQLJitResolveInfo", () => {
```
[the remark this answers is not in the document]
ruiaraujo: The only test I see using alias seems to be the `__typename` test.

ruiaraujo:
I am thinking whether we should make the `GraphQLJitResolveInfo` have an array instead of boolean true. It would be useful if we wanted to extend the computation including the arguments of the nodes without doing a breaking change.

boopathi:
An array of leaf types would be an option. If you have one other use-case in mind, it would be useful to base an assumption on and say that we need a leaf object instead of just an array. Array, like `true` would also lock down the API to mean similar things - it will be hard to extend an array than a leaf-object-type.

src/__tests__/resolve-info.test.ts
boopathi: It was harder to handle this specific case for unions where it behaves like an interface. It's like in GraphQL the composite types are implicitly part of a super interface like the following - 

```graphql
interface __Typename {
  __typename: String;
}
```

and the fact that a union implements this `__Typename` just messes around with the notion of a selection field inside of a selectionSet. If we get new internals, it's manual effort to handle it.

And I didn't see any value in including it as a part of the field selection for the use case we have.   The use-case of `if __typename field is requested; then do things differently` isn't important and we do not see such things.

src/resolve-info.ts
```diff
@@ -0,0 +1,247 @@
...
+ *   name: String
+ * }
+ * type Article implements Node {
+ *   id: ID!
+ *   title: String
+ * }
+ * union Card = User | Article
+ *
+ * - schema.getPossibleTypes(Card) would give [User, Article]
+ * - This function getPossibleTypes(schema, Card) would give [User, Article, Node]
```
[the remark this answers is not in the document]
boopathi: at the resolution level, we do not know in advance the resolved type already. But we might know about one of the overlapping interfaces and know at the time of resolution that it will be of that type. For example, a union of types that share a common interface.

src/resolve-info.ts
```diff
@@ -0,0 +1,247 @@
...
+
+type FragmentsType = GraphQLResolveInfo["fragments"];
+type GraphQLNamedOutputType = GraphQLNamedType & GraphQLOutputType;
+type GraphQLObjectLike = GraphQLInterfaceType | GraphQLObjectType;
+
+const memoizedGetReturnType = memoize2(getReturnType);
+const memoizedHasField = memoize2(hasField);
+const memoizedResolveEndType = memoize(resolveEndType);
+const memoizedGetPossibleTypes = memoize2(getPossibleTypes);
+const memoizedExpandFieldNodeType = memoize4(expandFieldNodeType);
```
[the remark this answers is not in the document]
boopathi: Aren't the fieldNodes generated and reused for all resolvers - same objects?

ruiaraujo:
I did not mean an array of leaf type, rather an array on the leafs instead of the boolean true.

src/resolve-info.ts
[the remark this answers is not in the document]
ruiaraujo: They are created per level in the query.
https://github.com/graphql/graphql-js/blob/6dfc94cc2aafcbf8727908dea4a99a8ec799da0f/src/execution/execute.js#L521

src/resolve-info.ts
ruiaraujo: According to @boopathi , they look the same.

src/resolve-info.ts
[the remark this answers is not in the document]
ruiaraujo: makes sense.

ruiaraujo:
:+1:

boopathi:
👍

PR #27, opened 2019-05-07
add an example with compiled persisted queries

boopathi:
Use a simple blog schema with mock data to showcase how to use compiled persisted queries with graphql-jit.


addityasingh:
👍

PR #28, opened 2019-05-08
add node 12; remove node 11

boopathi:
Also, remove installation of yarn as travis provides yarn by default.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,11 @@
 language: node_js
 node_js:
-- '11'
+- '12'
 - '10'
 cache:
+  yarn: true
   directories:
   - node_modules
-before_install:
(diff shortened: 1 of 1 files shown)


boopathi:
👍

src/__tests__/execution.test.ts
```diff
@@ -76,7 +76,7 @@ describe("Execute: Handles basic execution tasks", () => {
       [],
       []
     );
-    await Promise.resolve(); // For the promise to resolve
+    await new Promise(r => setImmediate(r)); // For the promise to resolve
```
boopathi: This is because the Promises execution is now back to how it was in Node 8 and is now an optimization in Node 12 - 1 microtask instead of 3. 

So, we wait longer to support both node 10 and 12.

PR #29, opened 2019-05-08
reexport types from resolve-info

boopathi:
For cases where you look-ahead two or three levels deep it's useful to have this to typecast.

+ `(info.fieldExpansion as FieldExpansion)["Foo"]`
+ `isLeafField(fieldExpansion)`


ruiaraujo:
:+1:

PR #30, opened 2019-05-08
Make compilation event-loop friendly.

ruiaraujo:
It will no longer hold the event loop for a large period of time at the cost of compilation throughput.

Before -  latest master
```
Results
{
    "type": "histogram",
    "min": 20628.938725,
    "max": 20628.938725,
    "sum": 20628.938725,
    "variance": null,
    "mean": 20628.938725,
    "std_dev": null,
    "count": 1,
    "median": 20628.938725,
    "p75": 20628.938725,
    "p95": 20628.938725,
    "p99": 20628.938725,
    "p999": 20628.938725
}
Total time: 20731.450504ms
```

After - this PR
```
Results
{
    "type": "histogram",
    "min": -0.9916180000000026,
    "max": 18.801353000000006,
    "sum": 110.67938100000002,
    "variance": 3.605352099798751,
    "mean": 0.4729888076923076,
    "std_dev": 1.8987764744168154,
    "count": 234,
    "median": 0.013759999999997774,
    "p75": 0.17402825000000277,
    "p95": 3.3057655000000032,
    "p99": 11.669598050000015,
    "p999": 18.801353000000006
}
Total time: 23558.98034ms
```

Given the single thread nature of NodeJS, this look to be a worthy compromise.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "graphql-jit",
-  "version": "0.1.3",
+  "version": "0.2.0",
   "description": "GraphQL JIT Compiler to JS",
   "main": "dist/index.js",
   "types": "./dist/index.d.ts",
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "graphql-tools": "^4.0.4",
     "jest": "^24.7.1",
     "lint-staged": "^8.1.5",
+    "metrics": "^0.1.21",
     "prettier": "^1.16.4",
     "ts-jest": "^24.0.2",
     "ts-node": "^8.0.3",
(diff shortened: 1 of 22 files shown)


vigneshshanmugam:
Not really sure if this comparison/benchmark makes sense in context of running the busy server handling requests. Can you at same time stress the server with load tests?

mcollina:
I would recommend against this, as it’s going to increase memory consumption in the real world case. In my experience, this _always_ result in worse performance in real life scenario.

The benchmark is not really measuring a real-world case. setImmediate is deferring execution to after the whole queue of events have been processed, which mean the server receives more requests at the same. If a query is “hot”, then it’s worthwhile compiling it synchronously as V8 can likely perform escape analysis and avoid several allocations.

ruiaraujo:
I am closing this for now since this creates more complexity.

PR #31, opened 2019-05-13
Move to a resolveInfo hook model.

ruiaraujo:
Not everyone will want to pay the extra cost for compiling the extensions to resolve info and this allows for consumers to extend their resolve info with other expensive tasks that can be moved to a compilation phase.

### DIFF
--- a/src/resolve-info.ts
+++ b/src/resolve-info.ts
@@ -21,13 +21,12 @@ import mergeWith from "lodash.mergewith";
 import { ObjectPath } from "./ast";
 import { memoize2, memoize4 } from "./memoize";
 
-export type ResolveInfoEnricher<T extends ResolveInfoEnricherInput> = (
-  i: ResolveInfoEnricherInput
-) => T | undefined;
+type NotGraphQLResolveInfo<T> = { [key in keyof T]: T[key] } &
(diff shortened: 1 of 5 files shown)


src/resolve-info.ts
```diff
@@ -61,51 +71,85 @@ export function isLeafField(obj: LeafField | FieldExpansion): obj is LeafField {
...
+  }: {
+    schema: GraphQLResolveInfo["schema"];
+    fragments: GraphQLResolveInfo["fragments"];
+    operation: GraphQLResolveInfo["operation"];
+    parentType: GraphQLObjectType;
+    fieldType: GraphQLOutputType;
+    fieldName: string;
+    fieldNodes: FieldNode[];
+  },
+  enricher?: (i: ResolveInfoEnricherInput) => any
```
boopathi: nit: returns `object | undefined` instead of `any`. And as we are spreading it below, `undefined` is actually fine.

ruiaraujo:
:+1:

ruiaraujo:
:+1:

boopathi:
👍

topicus:
👍

ruiaraujo:
:+1:

PR #35, opened 2019-06-11
Fix handling of variable to match spec behaviour.

### LINKED ISSUE #25: Filters not resolving properly?
Hi,

I'm having a bit of an issue with filters right now, which is confusing me.

I have a query which looks something like this:

```
query Foo($bar: String!) {
  system() {
    component(filter: { someKey: [$bar] }) {
     ...
  }
}
```

When I check the filter in the args passed to the resolvers, it's an empty object. If I change the filter to

```
    component(filter: { someKey: $bar }) {
```

I see the correct value in the filter.

Am I doing something wrong, or is there potentially a bug here?

Thanks.

ruiaraujo:
In practise this means that we now properly handle nested variable in objects and lists.

closes #25.

### DIFF
--- a/src/__tests__/inspect.test.ts
+++ b/src/__tests__/inspect.test.ts
@@ -2,9 +2,10 @@
  * Based on https://github.com/graphql/graphql-js/blob/master/src/jsutils/__tests__/inspect-test.js
  */
 
-import inspect, { nodejsCustomInspectSymbol } from "../inspect";
+import createInspect, { nodejsCustomInspectSymbol } from "../inspect";
 
 describe("inspect", () => {
+  const inspect = createInspect();
   it("undefined", () => {
     expect(inspect(undefined)).toEqual("undefined");
   });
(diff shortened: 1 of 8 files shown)


src/__tests__/inspect.test.ts
```diff
@@ -166,16 +165,37 @@ describe("inspect", () => {
...
+    expect(inspect([[new Foo()]])).toEqual("[[[Foo]]]");
 
     (Foo.prototype as any)[Symbol.toStringTag] = "Bar";
-    expect(inspect([[[new Foo()]]])).toEqual("[[[[Bar]]]]");
+    expect(inspect([[new Foo()]])).toEqual("[[[Bar]]]");
+  });
+});
+
+describe("createInspect", () => {
+  test("can increase the depth", () => {
```
ruiaraujo: This changes were required to allow having deeper inspection on the tests.

Also made the defaults match `graphql-js`.

ruiaraujo:
:+1:

ruiaraujo:
:+1:

boopathi:
👍

PR #42, opened 2019-06-18
feat: use operationName as the function name of the compiled output

### LINKED ISSUE #19: Name new Function to identify queries
Functions created in the executor are not named which makes hard to debug problematic queries.

- [ ] Add persisted query as functionName if possible
- [ ] Otherwise use the operationName

boopathi:
+ fix #19

### DIFF
--- a/src/execution.ts
+++ b/src/execution.ts
@@ -190,7 +190,15 @@ export function compileQuery(
   }`;
     const func = new Function(functionBody)();
     return {
-      query: createBoundQuery(context, document, func, getVariables),
+      query: createBoundQuery(
+        context,
+        document,
+        func,
+        getVariables,
+        context.operation.name != null
+          ? context.operation.name.value
+          : undefined
+      ),
       stringify
     };
   } catch (err) {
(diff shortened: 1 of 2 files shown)


ruiaraujo:
:+1:

src/execution.ts
```diff
@@ -212,87 +220,106 @@ export function createBoundQuery(
...
+   * we create an intermediate object with a method named as the
+   * intended function name. This is because Function.prototype.name
+   * is not writeable.
+   *
+   * http://www.ecma-international.org/ecma-262/6.0/#sec-method-definitions-runtime-semantics-propertydefinitionevaluation
+   *
+   * section: 14.3.9.3 - calls SetFunctionName
+   */
+  /* tslint:enable */
+  const ret = {
```
addityasingh: nit: I would use a more meaningful name such as `intermediateObject`. `ret` doesn't tells much about the intended purpose

boopathi: `ret` is a common variable name to indicate value used for **`ret`urn** . The comment above explains other things of using the object.

boopathi:
👍

PR #46, opened 2019-06-21
bugfix for using 'parent' as field name in SDL

wojtkiewicz:
Hi,
Currently there is a bug which prevents using string `parent` as a field name in GraphQL schema.

Example:
``` 
type MyType {
  parent: String
}
```

When querying for `parent` field there is an error:
`SyntaxError: Duplicate parameter name not allowed in this context`. 

I addressed this issue by introducing yet another global (`__parent`). Seems to be working fine.


boopathi:
👍

addityasingh:
👍

PR #48, opened 2019-06-28
Fix benchmarks (graphql-js returns different data than jit counterpart)

ajordanow:
graphql-js part returned 'Boolean cannot represent a non boolean value: "true"' error so the benchmark was not fair (the results are still favourable for jit though ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,13 +12,13 @@ code which yields much better performance.
 
 ```bash
 $ ts-node -T ./src/__benchmarks__/schema.benchmark.ts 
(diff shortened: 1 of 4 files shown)


ruiaraujo:
Ups, I remember fixing this at some point but I obviously forgot to follow up.

@ajordanow it also showcases that graphql-js degrades easily :sweat_smile:

ruiaraujo:
:+1:

addityasingh:
👍

Issue #52, opened 2019-07-16
Node configuration recommendations


wojtkiewicz:
Hi,
I'm running `graphql-jit` in production (as custom executor for Apollo GraphQL) and its performing worse than regular `graphql-js` implementation (higher cpu usage with similar throughput). I can see a difference in favor of `graphql-jit` when running benchmark (similar to `schema-many-resolvers.benchmark.ts`) but can't reproduce this in production setting. Results of `compileQuery` are cached properly between consecutive calls.

Do you have any recommendations for custom node config?
Do you have to tweak inlining options for v8 (like `max-inlined-bytecode-size`) for `graphql-jit` to work properly?

ruiaraujo:
which version of Nodejs/v8 are you using?

What kind of objects do your many resolvers return? Do they vary a lot in shapes, prototypes?
We are not tweaking v8.

wojtkiewicz:
I'm using v10.15 (LTS).

Queries are stable with respect to structure (parameterized, they hit compiled-query-cache all the time) but schema has a lot of optional fields that are nested in multidimensional arrays (common search result scenario where each record on a list has a lot of properties that can be quite complex). Typical response weights around 100-200kb.

I also noticed that `graphql-jit` is performing especially worse when app is querying for `__typename` extensively.

ruiaraujo:
Do you return a specific field for the typename or rely on the instance type to figure it out?

ruiaraujo:
@wojtkiewicz your use case seems quite interesting, if you could share a flamechart based on https://clinicjs.org/flame/ or something similar it would be easier to track down where it is deopting.

wojtkiewicz:
I don't have any specific resolvers for `__typename`. I have a bunch of flamegraphs from clinic already but I need to check what kind of stuff I can leak to the internet.

ruiaraujo:
@wojtkiewicz I found a regression that we introduced with #31. I also have another idea (inlining __typename) but it should be residual after #53.

wojtkiewicz:
@ruiaraujo that's a great news. I'm still working on a flamegraph because I'm afraid I need to make it look like we're running a cookie store. I should have something for tomorrow.

Apparently `__typename` field is queried by `apollo-ios` when you introduce conditional fragments to your queries. I'm expecting Android graphql implementations to do the same.

[...]

PR #53, opened 2019-07-17
Remove spreading from the enriching.

ruiaraujo:
This was a regression when we added the enriching functionality.
There is nothing conceptually that forces this to be expensive so move this to compiled code.

Before: `graphql-jit x 52,908 ops/sec ±3.06% (74 runs sampled)`
After: `graphql-jit x 92,752 ops/sec ±1.91% (76 runs sampled)`

Related with #52

### DIFF
--- a/src/resolve-info.ts
+++ b/src/resolve-info.ts
@@ -1,3 +1,4 @@
+import genFn from "generate-function";
 import {
   doTypesOverlap,
   FieldNode,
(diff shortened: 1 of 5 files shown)


src/resolve-info.ts
```diff
@@ -92,33 +92,61 @@ export function createResolveInfoThunk<T>(
...
+        parentType,
+        schema,
+        fragments,
+        operation
+      }) || {};
+    if (typeof enrichedInfo !== "object" || Array.isArray(enrichedInfo)) {
+      enrichedInfo = {};
+    }
+  }
+  let enrichedProperties = `return function getGraphQLResolveInfo(rootValue, variableValues, path) {
```
vigneshshanmugam: nit: try using `generate-function` instead?

ruiaraujo:
:+1:

boopathi:
👍

PR #56, opened 2019-07-29
benchmark with nested arrays

wojtkiewicz:
Hi,
This is related to #52.

I wrote a little benchmark to illustrate how both `graphql-js` and `graphql-jit` behave when you introduce arrays to your schema and fill them with data.

```
src/__benchmarks__/schema-many-resolvers.benchmark.ts:
graphql-js x 11,092 ops/sec ±6.37% (73 runs sampled)
graphql-jit x 80,470 ops/sec ±2.44% (77 runs sampled)

src/__benchmarks__/schema-nested-arrays.benchmark.ts:
graphql-js x 97.86 ops/sec ±10.78% (79 runs sampled)
graphql-jit x 602 ops/sec ±4.60% (77 runs sampled)
```

Main difference between two benchmarks is that in the `nested-arrays` case we have 25 articles attached to the author and each article has 25 badges and 25 adverts.

Not sure if these benchmark results are the best we can do. After all we have to resolve each element in an array so it degrades proportionally to an array length.

### DIFF
--- a/src/__benchmarks__/schema-nested-arrays.benchmark.ts
+++ b/src/__benchmarks__/schema-nested-arrays.benchmark.ts
@@ -1,0 +1,284 @@
+import Benchmark from "benchmark";
+import {
+  execute,
+  GraphQLBoolean,
+  GraphQLID,
+  GraphQLInt,
+  GraphQLList,
+  GraphQLNonNull,
+  GraphQLObjectType,
+  GraphQLSchema,
+  GraphQLString,
+  parse
+} from "graphql";
+import { compileQuery } from "../";
+
+const articlesCount = 25;
+const badgesCount = 25;
+const advertsCount = 25;
+
+const schema = getSchema();
+const document = parse(`
+{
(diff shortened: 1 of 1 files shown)


src/__benchmarks__/schema-nested-arrays.benchmark.ts
```diff
@@ -0,0 +1,277 @@
...
+            article(7),
+            article(8),
+            article(9),
+            article(10)
+          ])
+      }
+    }
+  });
+
+  for (let i = 0; i < 25; i++) {
```
ruiaraujo: Can we extract this 25 to a `const` at the top of the file to be easy to change the size?

wojtkiewicz: sure, sorry about that, done

ruiaraujo:
:+1:

angelaigreja:
:+1:

PR #57, opened 2019-07-29
Reuse arrays

ruiaraujo:
The intermediate array allocated to generate the final shape is pointless if the users do not plan on using the previous one.

This allows saving pointless array allocations.
It is disabled by default since it is an dangerous option for the cases where the user does require the original array to be kept.

Related with #52.
```
graphql-js x 111 ops/sec ±2.32% (72 runs sampled)
graphql-jit x 516 ops/sec ±4.42% (69 runs sampled)
graphql-jit - reuse arrays x 1,099 ops/sec ±2.52% (73 runs sampled)
```

### DIFF
--- a/src/execution.ts
+++ b/src/execution.ts
@@ -67,7 +67,7 @@ export interface CompilerOptions {
 
   // If set it reuses the memory of the arrays being transformed
   // saving the allocation costs of an intermediate array.
-  reuseArrays: boolean;
+  reuseArrayMemory: boolean;
 
   // Map of serializers to override
   // the key should be the name passed to the Scalar or Enum type
@@ -163,7 +163,7 @@ export function compileQuery(
       disablingCapturingStackErrors: false,
       customJSONSerializer: false,
       disableLeafSerialization: false,
-      reuseArrays: false,
+      reuseArrayMemory: false,
       customSerializers: {},
       ...partialOptions
     };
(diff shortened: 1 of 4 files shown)


src/execution.ts
```diff
@@ -940,14 +947,22 @@ function handleArrayValue(
...
  * @param {Iterable<any> | string} iterable possible iterable
  * @param {(a: any) => any} cb callback that receives the item being iterated
  * @returns {any[]} a new array with the result of the callback
  */
 function safeMap(
+  reuseArray: boolean,
   iterable: Iterable<any> | string,
   cb: (a: any, idx: number) => any
 ): any[] {
+  if (reuseArray && Array.isArray(iterable)) {
```
vigneshshanmugam: IMO, it feels like this optimization probably hurts more than giving improvements. 
https://v8.dev/blog/elements-kinds#avoid-elements-kind-transitions

Are you seeing benefits reusing the array?

ruiaraujo: Yes, the element kind will be same on the transition so there will not be any degradation.

I see double throughput on the nested arrays benchmark.

vigneshshanmugam: cool 👍

ruiaraujo:
I am not really sure whether we should do this. Even though it has positive effects, it is a super dangerous option that can introduce weird bugs depending on the requested fields.

In the other hands, it can help a lot specially in the case of lots of array allocations that would be wasteful.

PR #59, opened 2019-08-05
Support nested promises at any level in arrays.

ruiaraujo:
unpromisify had a special case to handle nested promises in an array so it could handle Array<Promise<any>>.

This change allows promises to be present anywhere.

### DIFF
--- a/src/execution.ts
+++ b/src/execution.ts
@@ -832,10 +832,11 @@ function compileListType(
   const listContext = createSubCompilationContext(context);
   // context depth will be mutated, so we cache the current value.
   const newDepth = ++listContext.depth;
+  const fieldType = type.ofType;
   const dataBody = compileType(
     listContext,
     parentType,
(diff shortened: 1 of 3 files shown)


ruiaraujo:
👍

src/execution.ts
```diff
@@ -1181,6 +1164,10 @@ export function isPromise(value: any): value is Promise<any> {
   );
 }
 
+export function isPromiseInliner(value: string): string {
+  return `${value} !== null && ${value} !== undefined && typeof ${value} === "object" && typeof ${value}.then === "function"`;
```
boopathi:
```suggestion
  return `${value} != null && typeof ${value} === "object" && typeof ${value}.then === "function"`;
```

boopathi:
👍

ruiaraujo:
:+1:

PR #60, opened 2019-08-09
Hoist all functions to the top level

ruiaraujo:
Currently, graphql-jit creates a lot of dynamic functions and takes advantages of closures to propagate context related with errors.

This creates additional memory pressure that could be avoided by hoisting the functions.
The PR introduces an execution context that is passed around to avoid the multiple closures well as
hoisting the grand majority of function created. The only ones left are the ones related with the abstract types which will be improved in a later PR.

Master
```
$ ts-node -T ./src/__benchmarks__/benchmarks.ts
Starting introspection
graphql-jit x 2,280 ops/sec ±5.12% (208 runs sampled)
Starting fewResolvers
graphql-jit x 97,191 ops/sec ±1.96% (222 runs sampled)
Starting manyResolvers
graphql-jit x 56,322 ops/sec ±4.60% (218 runs sampled)
Starting nestedArrays
graphql-jit x 485 ops/sec ±2.13% (219 runs sampled)
```

This PR
```
$ ts-node -T ./src/__benchmarks__/benchmarks.ts
Starting introspection
graphql-jit x 4,525 ops/sec ±1.04% (226 runs sampled)
Starting fewResolvers
graphql-jit x 121,675 ops/sec ±1.06% (221 runs sampled)
Starting manyResolvers
graphql-jit x 100,029 ops/sec ±1.38% (221 runs sampled)
Starting nestedArrays
graphql-jit x 915 ops/sec ±2.49% (222 runs sampled)
```

### DIFF
--- a/src/__tests__/abstract.test.ts
+++ b/src/__tests__/abstract.test.ts
@@ -13,12 +13,15 @@ import {
   GraphQLUnionType,
   parse
 } from "graphql";
-import { compileQuery } from "../index";
+import { compileQuery, isCompiledQuery } from "../index";
 
 function graphql(schema: GraphQLSchema, query: string) {
-  const ast = parse(query);
-  const compiled: any = compileQuery(schema, ast, "");
-  return compiled.query(undefined, undefined, {});
+  const document = parse(query);
+  const prepared = compileQuery(schema, document, "");
+  if (!isCompiledQuery(prepared)) {
+    return prepared;
+  }
+  return prepared.query(undefined, undefined, undefined);
 }
 
 class Dog {
--- a/src/execution.ts
+++ b/src/execution.ts
@@ -108,14 +108,8 @@ interface DeferredField {
   args: Arguments;
 }
 
-export type SuccessCallback = (d: object | null) => void;
-export type ErrorCallback = (e: Error | null) => void;
-
-export type JITCallback = (
-  p: object,
-  d: object | null,
-  e: Error | null
-) => void;
+export type SuccessCallback = (parent: any, d: object) => void;
+export type ErrorCallback = (e: Error) => void;
 
 export interface CompiledQuery {
   operationName?: string;
@@ -297,7 +291,7 @@ export function createBoundQuery(
       if (operation === "mutation") {
         const serial = serialPromiseExecutor(callback, globalErrorHandler);
         executor = serial.addToQueue;
-        resolveIfDone = serial.startExecution;
+        resolveIfDone = serial.startOrContinueExecution;
       } else {
         const loose = loosePromiseExecutor(callback, globalErrorHandler);
         executor = loose.executor;
@@ -381,7 +375,10 @@ function compileOperation(context: CompilationContext) {
   );
   let body = generateUniqueDeclarations(context, true);
   body += `const ${GLOBAL_DATA_NAME} = ${topLevel}\n`;
-  body += compileDeferredFields(context);
+  body += compileDeferredFields(
+    context,
+    context.operation.operation === "mutation"
+  );
   return body;
 }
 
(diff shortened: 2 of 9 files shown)


src/execution.ts
```diff
@@ -1601,3 +1547,14 @@ function normalizeError(err: Error): GraphQLError {
 function isInvalid(value: any): boolean {
   return value === undefined || value !== value;
 }
+
+function getParentIndexes(context: CompilationContext) {
+  let args = "";
+  for (let i = 1; i <= context.depth; ++i) {
+    if (i > 1) {
```
ruiaraujo: Add comment for this.

src/execution.ts
```diff
@@ -361,6 +367,7 @@ function postProcessResult(
  */
 function compileOperation(context: CompilationContext) {
   const type = getOperationRootType(context.schema, context.operation);
+  const serialExecution = context.operation.operation === "mutation";
```
topicus: Should we create an enum with operation names?

src/execution.ts
```diff
@@ -1601,3 +1547,14 @@ function normalizeError(err: Error): GraphQLError {
 function isInvalid(value: any): boolean {
   return value === undefined || value !== value;
 }
+
+function getParentIndexes(context: CompilationContext) {
```
topicus: this should be name like: getParentArgIndexes.

src/execution.ts
```diff
@@ -395,83 +444,135 @@ function compileOperation(context: CompilationContext) {
...
+    [`${GLOBAL_PARENT_NAME}.${name}`],
+    responsePath
+  );
+  const parentIndexes = getParentIndexes(context);
+  const resolverName = getResolverName(parentType.name, fieldName);
+  const resolverHandler = getHoistedFunctionName(
+    context,
+    `${name}${resolverName}Handler`
+  );
+  const topLevelArgs = getArgumentsVarName(resolverName);
```
addityasingh: What does `getArgumentsVarName` do? Arguments with Var is confusing to me

src/execution.ts
```diff
@@ -395,83 +444,135 @@ function compileOperation(context: CompilationContext) {
...
+  const executionInfo = getExecutionInfo(
+    subContext,
+    parentType,
+    fieldType,
+    fieldName,
+    fieldNodes,
+    responsePath
+  );
+  const emptyError = createErrorObject(context, fieldNodes, responsePath, '""');
+  const resolverParentObject = originPaths.join(".");
```
addityasingh: Since the result of `.join()` is a string, I would use `resolverParent` or `resolverParentPath`

src/execution.ts
ruiaraujo: This is the only place where this comparison is made so I do not see the need.

src/execution.ts
```diff
@@ -77,25 +85,67 @@ export interface CompilerOptions {
  *
  * It stores deferred nodes to be processed later as well as the function arguments to be bounded at top level
  */
-interface CompilationContext extends ExecutionContext {
-  dependencies: Map<string, (...args: any[]) => any>;
+interface CompilationContext extends GraphQLContext {
+  resolvers: { [key: string]: GraphQLFieldResolver<any, any, any> };
```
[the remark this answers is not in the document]
ruiaraujo: Internally v8 will use Maps if there are too many values.
If there are few, having an internal shape will provide faster property access.

src/execution.ts
ruiaraujo: Creates the variable name for the arguments input for the resolver. I always accept better names. :smile:

addityasingh: I would just use `getArgumentsName`

src/__tests__/abstract.test.ts
```diff
@@ -13,12 +13,15 @@ import {
...
-import { compileQuery } from "../index";
+import { compileQuery, isCompiledQuery } from "../index";
 
 function graphql(schema: GraphQLSchema, query: string) {
-  const ast = parse(query);
-  const compiled: any = compileQuery(schema, ast, "");
-  return compiled.query(undefined, undefined, {});
+  const document = parse(query);
+  const prepared = compileQuery(schema, document, "");
+  if (!isCompiledQuery(prepared)) {
```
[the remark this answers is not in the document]
ruiaraujo: Better errors when tests are broken.

ruiaraujo:
:+1:

addityasingh:
👍

PR #61, opened 2019-08-15
Rework benchmarks.

ruiaraujo:
The logic to run the benchmark is always the same and usually we want to see the effects on all of them. Furthermore, variables are highly common and JSON transformation so these are included in the microbenchmarks.

Add time based benchmark that can be used for profilling.


ruiaraujo:
:+1:

angelaigreja:
:+1:

PR #62, opened 2019-09-03
Fix list regression with nested promises.

ruiaraujo:
Now the generated code mirrors the style of the resolvers handling.


ruiaraujo:
:+1:

oporkka:
:+1:

PR #64, opened 2019-09-16
Remove functionBody unless requested.

ruiaraujo:
With large amount of persisted queries, it has a noticeable memory cost.


ruiaraujo:
👍

addityasingh:
👍

Issue #67, opened 2020-01-06
Apollo server integration


guigaoliveira:
Hi, is it possible to integrate with apollo server?

wojtkiewicz:
Yes, I did that successfully on production.

You can provide custom `executor` using `ApolloServerConfig` but to be honest this looks like a pretty repeatable code. 

@ruiaraujo I think I can contribute PR to this repo with such integration if you want.

guigaoliveira:
Interesting, I would like to see some example, looks promising :)

ruiaraujo:
@wojtkiewicz yes please.We only use Apollo tools at Zalando so a tried contribution would be very valuable.

Taraluktus:
Would also love to see an example of it.

wojtkiewicz:
@ruiaraujo Sorry to have kept you waiting. I had few side hustles to complete. I will contribute something this week.

OliverCole:
Paging @benawad who may wish to add this to https://github.com/benawad/node-graphql-benchmarks when published.

PR #72, opened 2020-01-27
added sample integration between graphql-jit and apollo

wojtkiewicz:
Related to: https://github.com/zalando-incubator/graphql-jit/issues/67

This PR contains sample integration between graphql-jit and Apollo Server. It can be used as a starting point to build more robust `apollo-server-graphql-jit` component in the future. What you probably want is to add metrics and logs for instance.

For legal reasons I can't contribute entire module that I was running on production (at least not yet, it has to pass lengthy security clearance to be open sourced).

### DIFF
--- a/examples/blog-apollo-server/package.json
+++ b/examples/blog-apollo-server/package.json
@@ -1,0 +1,9 @@
+{
+  "name": "blog-apollo-server",
+  "version": "0.1.0",
+  "dependencies": {
+    "apollo-server": "^2.9.16",
+    "graphql-jit": "^0.4.2",
+    "tiny-lru": "^7.0.2"
+  }
+}
--- a/examples/blog-apollo-server/package.json
+++ b/examples/blog-apollo-server/package.json
@@ -3,7 +3,6 @@
   "version": "0.1.0",
   "dependencies": {
     "apollo-server": "^2.9.16",
-    "graphql-jit": "^0.4.2",
     "tiny-lru": "^7.0.2"
   }
 }
--- a/examples/blog-apollo-server/src/executor.ts
+++ b/examples/blog-apollo-server/src/executor.ts
@@ -1,0 +1,33 @@
+import { GraphQLExecutor } from "apollo-server-core";
+import { GraphQLSchema } from "graphql";
+import {
+  CompiledQuery,
+  compileQuery,
+  CompilerOptions,
+  isCompiledQuery
+} from "graphql-jit";
+import LRU from "tiny-lru";
+
+export const executor = (
+  schema: GraphQLSchema,
+  cacheSize = 1024,
+  compilerOpts: Partial<CompilerOptions> = {}
+): GraphQLExecutor => {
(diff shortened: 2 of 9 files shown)


examples/blog-apollo-server/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "blog-apollo-server",
+  "version": "0.1.0",
+  "dependencies": {
+    "apollo-server": "^2.9.16",
+    "graphql-jit": "^0.4.2",
```
boopathi: point to the GraphQL-jit from `../../` (or) remove this dependency and replace imports to use `../../`

examples/blog-apollo-server/src/executor.ts
```diff
@@ -0,0 +1,33 @@
...
+  cacheSize = 1024,
+  compilerOpts: Partial<CompilerOptions> = {}
+): GraphQLExecutor => {
+  const cache = LRU<CompiledQuery>(cacheSize);
+  return async ({ context, document, operationName, request, queryHash }) => {
+    const prefix = operationName || "NotParametrized";
+    const cacheKey = `${prefix}-${queryHash}`;
+    let compiledQuery = cache.get(cacheKey);
+    if (!compiledQuery) {
+      const compilationResult = compileQuery(schema, document, operationName || undefined, compilerOpts);
```
boopathi:
```suggestion
      const compilationResult = compileQuery(schema, document, operationName, compilerOpts);
```

Is there a valid case where ` || undefined` is useful?

examples/blog-apollo-server/package.json
wojtkiewicz: Done. I had to upgrade `graphql` to `^14.5.3` for this code to compile though. Is this ok with you?

examples/blog-apollo-server/src/executor.ts
wojtkiewicz: Unfortunately apollo defines operationName as `string | null` and graphql-jit requires `string | undefined`. So I need this extra `null => undefined` conversion to compile.

examples/blog-apollo-server/package.json
ruiaraujo: The upgrade is okay, but I wonder why it was required?

examples/blog-apollo-server/src/executor.ts
ruiaraujo: @boopathi we should extend what we expect. We dont care if it is null or undefined.

examples/blog-apollo-server/package.json
wojtkiewicz: Main culprit in this case is `GraphQLSchema`. Graphql-jit was expecting older model and apparently there are some differences that compiler detected. Downgrading `apollo-server` should also help if you need more testing with latest `graphql-js`.

ruiaraujo:
:+1:

boopathi:
👍

PR #73, opened 2020-02-06
README: Update benchmark command

IvanGoncharov:
Without `NODE_ENV=production` `graphql-js` uses dev version of `instanceOf`:
https://github.com/graphql/graphql-js/blob/master/src/jsutils/instanceOf.js


boopathi:
👍

addityasingh:
👍

PR #77, opened 2020-03-25
fix: allow js keywords as field names

### LINKED ISSUE #76: Unexpected token 'import'
We would like to switch from [express-graphql](https://github.com/graphql/express-graphql) to [express-gql](https://github.com/benawad/express-gql) (which uses graphql-jit) and are having trouble with one of our queries: `import(id: Int!): Import`

The function `compileQuery(...)` returns the following error:
```
{"errors":[{"message":"Unexpected token 'import'"}]}
```

As far as I understand, graphql-jit compiles the query into a JS function and `import` is a reserved keyword in JavaScript but not in GraphQL.

Is there any way to modify the function generation so that those keywords are okay to use?

boopathi:
+ fix #76 

Changes the fieldName variable name in the hoisted functions by prefixing the fieldName with `__field`

### DIFF
--- a/src/__tests__/execution.test.ts
+++ b/src/__tests__/execution.test.ts
@@ -16,6 +16,7 @@ import {
   parse
 } from "graphql";
 import { CompiledQuery, compileQuery, isCompiledQuery } from "../execution";
+import { makeExecutableSchema } from "graphql-tools";
 
 function executeArgs(args: any) {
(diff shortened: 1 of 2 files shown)


ruiaraujo:
:+1:

boopathi:
👍

PR #80, opened 2020-04-07
Upgrade GraphQL to v15

frikille:
Fixed the two failing tests caused by text change and order of object properties.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "@types/node": "^10.11.7",
     "benchmark": "^2.1.4",
     "codecov": "^3.3.0",
-    "graphql": "^14.5.3",
+    "graphql": "^15.0.0",
     "graphql-tools": "^4.0.4",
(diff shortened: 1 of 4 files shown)


ruiaraujo:
For some reason yarn is complaining. :confused: 

Can you run it and commit what is wrong?

addityasingh:
@frikille You need to update the yarn.lock file by running yarn again. The build fails with below error

>yarn --frozen-lockfile
yarn install v1.15.2
[1/4] Resolving packages...
error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.

frikille:
@addityasingh Updated `yarn.lock` and it seems CI is passing now

addityasingh:
👍

ruiaraujo:
:+1:

PR #82, opened 2020-05-17
Fix compilation and execution for circular dependencies in input types

### LINKED ISSUE #81: Request variables with circular dependency
Hi,

I encountered this issue when making a create request for entity A that has a circular reference with B. Schema definition is implemented with type-graphql roughly like this.
```
@ObjectType()
@InputType('FooInput')
class Foo {
    @Field(() => Number)
    id: number;
    @Field(() => Bar, {nullable: true})
    bar: Bar;
}

@ObjectType()
@InputType('BarInput')
class Bar {
    @Field(() => Number)
    id: number;
    @Field(() => Foo, {nullable: true})
    foo: Foo;
}

@Resolver(Foo)
class FooResolver {
    @Mutation()
    addFoo(@Arg('foo') foo: Foo): string {
        return 'ok';
    }
}
```

Now if I try to send a query
```
mutation AddFoo($foo: FooInput!){
    addFoo(foo: $foo)
}
```
with variables
```
{
  "foo": {
    "id": 1,
    "bar": null
  }
}
```
the query execution gets stuck in a recursive loop during `compileQuery` call or more specifically in [generateInput](https://github.com/zalando-incubator/graphql-jit/blob/master/src/variables.ts#L137) function. Apparently when going through the request variables in `if (isInputType(varType))` block generating another call to itself and this goes on until out of memory error.

I'm using apollo-server with custom executor like described in the examples. Similar issues do not occur with default apollo-server config or by running the query with with graphql-js execute.
```
new ApolloServer({
    executor: ({ source }) => {
      const compiled = compileQuery(schema, parse(source)); // <-- never returns from here
      // ...
    }
});
```

boopathi:
+ fix #81 

When input types are recursive in nature, the compilation depended on generating it inline. This PR hoists it to a function and generates it only once for input object types. This way, instead of relying on inlined object access, we have a getter and a setter defined at the top of the generated code - `getPath` and `setPath` - this moves the object access (the path - `input["var"]["foo"]`) to runtime - via getPath and setPath - so as to pass different paths as we move down the tree.

Also, we compute the circular dependencies at runtime and throw them as separate user errors/ For this we now have the `Set` of visited inputs captured at the top-level. At the time of this writing, I've not handled the false positives where - `var1.foo` is the same object as `var2.foo` - which does not form a circular dependency. This is unintended anyway because variables are generally passed down via request and is most likely deserialized from JSON. A JSON.parse cannot result in such cases anyway.

### DIFF
--- a/src/__tests__/recursive-input.test.ts
+++ b/src/__tests__/recursive-input.test.ts
@@ -1,0 +1,357 @@
+import { DocumentNode, GraphQLSchema, parse, validate } from "graphql";
+import { makeExecutableSchema } from "graphql-tools";
+import { compileQuery, isCompiledQuery } from "../execution";
+
+describe("recursive input types", () => {
+  describe("simple recursive input", () => {
+    const schema = makeExecutableSchema({
+      typeDefs: `
+        type Query {
+          foo(input: FooInput): String
+        }
+        input FooInput {
+          foo: FooInput
+        }
+      `,
+      resolvers: {
+        Query: {
+          foo(_, args) {
+            // used as the acutal value in test matchers
+            return JSON.stringify(args);
+          }
+        }
+      }
+    });
+
+    test("should not fail for recursive input without variables", () => {
+      const query = parse(`
+        {
+          foo(input: {
+            foo: {
+              foo: {
+                foo: {
+                  foo: {}
+                }
+              }
+            }
+          })
+        }
+      `);
+
+      const result = executeQuery(schema, query);
+
+      expect(result.errors).toBeUndefined();
+      expect(result.data.foo).toBe(
+        JSON.stringify({
+          input: {
+            foo: { foo: { foo: { foo: {} } } }
+          }
+        })
+      );
+    });
+
+    test("should not fail with variables using recursive input types", () => {
+      const document = parse(`
+        query ($f: FooInput) {
+          foo(input: $f)
+        }
+      `);
+      const variables = {
+        f: {
+          foo: { foo: { foo: {} } }
+        }
+      };
+
+      const result = executeQuery(schema, document, variables);
+      expect(result.errors).toBeUndefined();
+      expect(result.data.foo).toBe(
+        JSON.stringify({
+          input: { foo: { foo: { foo: {} } } }
+        })
+      );
+    });
+
+    // when the recursive variable appers at a nested level
+    test("should not fail with variables using recursive input types - 2", () => {
+      const document = parse(`
+        query ($f: FooInput) {
+          foo(input: {
+            foo: { foo: { foo: $f } }
+          })
+        }
+      `);
+      const variables = {
+        f: {
+          foo: { foo: { foo: {} } }
+        }
+      };
+
+      const result = executeQuery(schema, document, variables);
+      expect(result.errors).toBeUndefined();
+      expect(result.data.foo).toBe(
+        JSON.stringify({
+          input: { foo: { foo: { foo: { foo: { foo: { foo: {} } } } } } }
+        })
+      );
+    });
+
+    test("should work with multiple variables using the same recursive input type", () => {
+      const document = parse(`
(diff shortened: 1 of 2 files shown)


src/__tests__/recursive-input.test.ts
```diff
@@ -0,0 +1,357 @@
...
+        }
+        input OrFilter {
+          left: Filter
+          right: Filter
+        }
+      `,
+      resolvers: {
+        Query: {
+          products(_, args) {
+            // used as the acutal value in test matchers
```
ruiaraujo, line 201:
```suggestion
            // used as the actual value in test matchers
```

src/__tests__/recursive-input.test.ts
```diff
@@ -0,0 +1,357 @@
...
+          foo(input: FooInput): String
+        }
+        input FooInput {
+          foo: FooInput
+        }
+      `,
+      resolvers: {
+        Query: {
+          foo(_, args) {
+            // used as the acutal value in test matchers
```
ruiaraujo, line 19:
```suggestion
            // used as the actual value in test matchers
```

src/variables.ts
```diff
@@ -98,7 +105,15 @@ export function compileVariableParsing(
...
     context.responsePath = addPath(context.responsePath, varName);
-    mainBody += generateInput(context, varType, varName, hasValueName, false);
+    mainBody += generateInput(
+      context,
+      varType,
+      varName,
+      hasValueName,
+      false,
+      false,
+      false
```
ruiaraujo, line 115: At this stage we should consider an object input to have named properties.

src/__tests__/recursive-input.test.ts
```diff
@@ -0,0 +1,588 @@
+import { DocumentNode, GraphQLSchema, parse, validate } from "graphql";
+import { makeExecutableSchema } from "graphql-tools";
```
ardatan, line 2: You can use `@graphql-tools/schema` instead because npm package `graphql-tools` has been deprecated :)

jgoux:
Hello, is there anything particular blocking this PR? It would unlock a lot of people if it can be merged. 🤞

tuan231195:
@boopathi @ruiaraujo, is there any update about this PR? It would solve lots of issues for us

mnlbox:
Any plan to finalize and merge this PR ??

Urigo:
@boopathi @ruiaraujo What can we do in order to help this land?
Should we open another PR with a different approach?

oporkka:
> @boopathi Thanks for this awesome project but this issue and the PR have been open for 4 years. I'm trying to figure out what is the blocker here because we would like to help to get this issue resolved.
> 
> How we can help to proceed with the solution and fix this issue? This issue is blocking us from using GraphQL-Mesh as it uses graphQL-jit behind the scenes. We would like to connect Hasura to GraphQL-Mesh too and as we tried in a POC, we have this circular dependency issue there as well. So we are glad to contribute and help on this issue to get it resolved.

@BabakScript I think what is pending, is the missing benchmarks that @ruiaraujo was mentioning earlier.

> I would love to have some benchmarks that stress query execution with highly nested variables (like the filter test in this PR) between master and this PR.
> 
> If this is punishing for the before case, maybe we can consider detecting the circular case in the types and bail out to such an implementation while relying on inline access for the normal case.

@BabakScript would you be able to contribute the above benchmarks?

boopathi:
Some benchmarks are available here in these PRs - #228 and #229 and it's not performant and I was trying to optimize it. I'll get back to this when I find some time.

BabakScript:
> @BabakScript would you be able to contribute the above benchmarks?

@oporkka I guess as @boopathi mentioned the benchmarks part is somehow done. We didn't contribute to this project yet but we can try to read the code and check if we can help with the optimization part.


BTW, Thanks @boopathi  for the update. I guess #228 is also a promising and good approach.

PR #84, opened 2020-06-08
Make fast-json-stringify optional

### LINKED ISSUE #83: Make fast-json-stringify optional
Thanks for the great project.

**Is your feature request related to a problem? Please describe.**

I'm experimenting with using `graphql-jit` in the browser. Currently `graphql-jit` is bundled with `fast-json-stringify`, which makes its size extremely big to be used in browser.

https://bundlephobia.com/result?p=graphql-jit

Removing `fast-json-stringify` would cut the size in half.

**Describe the solution you'd like**

Currently, we have the option `compilerOptions.customJSONSerializer`, which accepts a boolean value. However the option does not provide much use because it only switches between `JSON.stringify` and `fastJSON`, not really 'custom'.

My proposal is that we replace it with `compilerOptions.fastJSON`, which will either be `undefined` or a user-supplied `require('fast-json-stringify')`.

hoangvvo:
Close #83 

This is still up to discussion and may be closed if it deems likewise. This is a breaking change

```
const fastJson = require('fast-json-stringify')
const { compileQuery } = require("graphql-jit");

const compiledQuery = compileQuery(schema, document, { fastJson });
```

Caveat: 
- This may be confusing for end-user.
- `queryToJSONSchema` is tightly coupled with `fast-json-stringify` ([`nullable` prop](https://github.com/fastify/fast-json-stringify#nullable))
- May break if `fast-json-stringify` has breaking change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ Compiles the `document` AST, using an optional operationName and compiler option
     so this option should only be set to true if there are strong assurances that the values are valid.
   - `customSerializers` {Object as Map, default: {}} - Replace serializer functions for specific types. Can be used as a safer alternative
     for overly expensive serializers
-  - `customJSONSerializer` {boolean, default: false} - Whether to produce also a JSON serializer function using `fast-json-stringify`. The default stringifier function is `JSON.stringify`
+  - `fastJson` {`require(fast-json-stringify)`, default: undefined} - Whether to produce also a JSON serializer function using `fast-json-stringify`. The default stringifier function is `JSON.stringify`
 
 #### compiledQuery.compiled(root: any, context: any, variables: Maybe<{ [key: string]: any }>)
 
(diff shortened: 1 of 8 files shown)


ruiaraujo:
I will try reviewing this weekend.

ruiaraujo:
I would prefer to extract the json to a separate package. I am not a fan of having it integrated since there are very limited advantages to having it included.

It would also be natural to expose graphql query to binary.

@boopathi wdyt?

boopathi:
Yes. I also agree with the caveats mentioned in the description. I think we can start with this and think about a model after we start including serializing to other formats like binary.

hoangvvo:
I agree to this. I'm actually ok with not even having that separate package in the meantime.

We can still keep `customJSONSerializer`, but letting it be a function that accepts `context` from `buildCompilationContext` and returns a custom stringify function.

```
customJSONSerializer: (context) => {
  jsonSchema = outOwnQueryToJSONSchema(context);
  return fastJson(jsonSchema)
}
```

Closing.

PR #85, opened 2020-06-13
Upgrade TypeScript to 3.9.5 (fixing a tiny bug in the process)

airhorns:
While working on a different issue, I noticed `graphql-jit` didn't typecheck on newer TypeScripts. This bumps the compiler version used in development and fixes two small issues that arose from the bump. See the commits for a description of each!

PR #86, opened 2020-06-13
GraphQL >= 15.1.0 types fix

airhorns:
This is on top of #85 to keep things simple happy to rebase if they shouldn't share a destiny :)

This fixes an issue where if using `graphql-jit` with `graphql` >= 15.1.0 the included types no longer typechecked.

graphql moved the Maybe utility internally from the default export of `graphql/tsutils/maybe` to a named export of `graphql/jsutils/maybe`. This breaks `graphql-jit`s shipped types when using a new `graphql` version where Maybe has moved: 

```
src/execution.ts:36:19 - error TS2307: Cannot find module 'graphql/tsutils/Maybe' or its corresponding type declarations.

36 import Maybe from "graphql/tsutils/Maybe";
                     ~~~~~~~~~~~~~~~~~~~~~~~
```

We could just update the paths to reference the new location, but then `graphql-jit` would strictly require graphql>=15.1.0 to typecheck as old graphqls wouldn't have anything at the new path. This requirement seems kind of silly just to fix a small typescript issue. In order to support both old and new versions of `graphql`'s types, I think it's easiest to just move the type inside this repo since it's really simple. TypeScript will find it equivalent to the one from graphql, and we can import it from the same path regardless of graphql version.


boopathi:
>  In order to support both old and new versions of graphql's types, I think it's easiest to just move the type inside this repo since it's really simple.

Yes. I agree. Thank you. Can you please rebase with master? The PR looks good.

ruiaraujo:
@airhorns can you run `yarn lint-fix`. The build is not passing.

airhorns:
Ok, build is passing now! I switched from `graphql-tools` to `@graphql-tools/schema` to avoid bringing in a huge number of new development dependencies, they switched to a sub-packages style where the top level `graphql-tools` package imports all the sub packages, which were importing all of babel and some ESM modules that jest didn't like. `@graphql-tools/schema` has a much smaller dependency list and has support for the new `graphql-js` version that prompted all this in the first place.

ruiaraujo:
:+1:

OtaK:
Any updates on this PR? 
It prevents using `graphql-jit` with the updated parts of the ecosystem like recent versions of `graphql`, `type-graphql` etc.

airhorns:
@ruiaraujo it'd be awesome to get this in!

ruiaraujo:
@boopathi can you take a look ?

boopathi:
👍

PR #88, opened 2020-07-19
Remove fast-json-stringify and allows customJSONSerializer to be a function

### LINKED ISSUE #83: Make fast-json-stringify optional

hoangvvo:
This PR allows `options.customJSONSerializer` to be a function to be called with `CompilationContext` to returns a serializer function. It will be a breaking change.

Something like `queryToJSONSchema` in [json.ts](https://github.com/zalando-incubator/graphql-jit/blob/master/src/json.ts) is not a concern of this package and should be reimplemented in userland.

```js
import { queryToJSONSchema } from '../user-land/json';

compileQuery(schema, document, "", {
  customJSONSerializer: (context) => {
    const jsonSchema = queryToJSONSchema(context);
    return fastJson(jsonSchema);
  }
});
```

### Why

- `fast-json-stringify` is way too large if `graphql-jit` is ever used in browser.
- We pin `fast-json-stringify` at v1 which misses out a lot of features from v2. User should get to choose what JSON serializer to use.

Fix #83


hoangvvo:
I tried to publish it as a fork. Bundle size went down from `226.9 kb` -> `70.3 kb` (~70%)

https://bundlephobia.com/result?p=@hoangvvo/graphql-jit@0.5.0

compared to 

https://bundlephobia.com/result?p=graphql-jit@0.4.3

README.md
```diff
@@ -110,7 +110,7 @@ Compiles the `document` AST, using an optional operationName and compiler option
     so this option should only be set to true if there are strong assurances that the values are valid.
   - `customSerializers` {Object as Map, default: {}} - Replace serializer functions for specific types. Can be used as a safer alternative
     for overly expensive serializers
-  - `customJSONSerializer` {boolean, default: false} - Whether to produce also a JSON serializer function using `fast-json-stringify`. The default stringifier function is `JSON.stringify`
+  - `customJSONSerializer` {function, default: false} - A function to be called with [`CompilationContext`](https://github.com/zalando-incubator/graphql-jit/blob/master/src/execution.ts#L87) to produce also a JSON serializer function. The default stringifier function is `JSON.stringify`
```
ruiaraujo, line 113:
```suggestion
  - `customJSONSerializer` {function, default: undefined} - A function to be called with [`CompilationContext`](https://github.com/zalando-incubator/graphql-jit/blob/master/src/execution.ts#L87) to produce also a JSON serializer function. The default stringifier function is `JSON.stringify`
```

ruiaraujo:
Isn't this equivalent to removing the feature fully?

README.md
hoangvvo: In code, the default value is `false`. Is this okay to write it as `undefined` in the doc (even though it does not matter actually)

https://github.com/zalando-incubator/graphql-jit/blob/8b678a10b9ec1e16db84b0566f4d64fe9be9b5de/src/execution.ts#L214

hoangvvo:
Sorry for the late response. It is helpful to me, but I'm not sure if others are depending on it (in which case let's drop this PR).

By the name `customJSONSerializer`, the option should allow actually "custom" JSON serializer, not specifically the built in one.

Pretty much my reasons have already been mentioned above in this PR. This feature seems to be "out of the scope" of this package.

ruiaraujo:
@boopathi what do you think about this? We could move to a mutiple package setup and add the JSON stringifier with the idea of later adding a protobuf one for example.

boopathi:
yes I think that's a good idea. we can add and move a few other things as well.

hoangvvo:
Are we still interested in proceeding in this direction? I would love to reopen another PR after discussing the good approach forward. Thanks!

Issue #92, opened 2020-10-03
Error metadata from custom directives


aldis-ameriks:
I'm looking into using custom directives for validation using graphql-tools based on the docs here.
https://www.graphql-tools.com/docs/schema-directives/#enforcing-value-restrictions

The validation seems to be working and the error message in the compiled query result matches the one which is thrown from the validation directive. However, I cannot seem to get the originalError, including extensions from the error that's thrown from the validation directive.

I saw this issue #2 which might be similar, but I'm not actually throwing the error from within the resolver, so it may be a different case.

I've created a sandbox with the example code to illustrate what I'm trying to achieve and how.
https://codesandbox.io/s/long-grass-hlul0?file=/src/index.js

Is this type of case supported and/or is there some way to get the error metadata?
Thanks!

aldis-ameriks:
Using v0.5.0 I'm able to retrieve the error metadata from input validation errors using custom directives / scalars.

PR #93, opened 2020-10-10
Preserve original error

aldis-ameriks:
This is an attempt to address issue #92.
After going through the codebase, I managed to find where the `originalError` is lost in my use case, when an error is thrown from custom directive.
After these changes, I'm able to preserve the `originalError`.


boopathi (approved):

ruiaraujo:
:+1:

boopathi:
👍

PR #94, opened 2020-10-25
Implement subscription

### LINKED ISSUE #74: Support createSourceEventStream?
Wonderful package, lots to learn from, thank you for open sourcing it!

Any plans to support `createSourceEventStream` for subscriptions? Any tips on how I'd go about PRing it?

hoangvvo:
This implements subscription support via `jit.subscribe`.

Close #74


hoangvvo:
As per https://github.com/zalando-incubator/graphql-jit/issues/74#issuecomment-686740433, we have one failing test:

```
Subscription Initialization Phase › resolves to an error for source event stream resolver errors

    expect(received).toEqual(expected)

    Difference:

    - Expected
    + Received

      Object {
        "errors": Array [
    -     Object {
    -       "locations": Array [
    -         Object {
    -           "column": 13,
    -           "line": 3,
    -         },
    -       ],
    -       "message": "test error",
    -       "path": Array [
    -         "importantEmail",
    -       ],
    -     },
    +     [GraphQLError: test error],
        ],
      }
```

ruiaraujo:
@hoangvvo apologies for not reacting earlier.

Do you plan to finish this PR?

hoangvvo:
> @hoangvvo apologies for not reacting earlier.
> 
> Do you plan to finish this PR?

Hey, no worry. I do plan to finish the PR. I will get back to this once I have a chance sometimes later this week.

src/execution.ts
```diff
@@ -227,17 +241,29 @@ export function compileQuery(
...
+      new Function("return " + functionBody)(),
+      getVariables,
+      context.operation.name != null ? context.operation.name.value : undefined
+    );
+
+    // Subscription
     const compiledQuery: InternalCompiledQuery = {
-      query: createBoundQuery(
+      query,
+      subscribe: createBoundSubscribe(
```
ruiaraujo, line 263: This property should only be available when the query is a subscription.

hoangvvo:
@ruiaraujo For the failing test, if I use `jest.toMatchObject` instead of `jest.toEqual`, it shows the error more clearly:

```
Subscription Initialization Phase › resolves to an error for source event stream resolver errors

    expect(received).toMatchObject(expected)

    Expected value to match object:
      {"errors": [{"locations": [{"column": 13, "line": 3}], "message": "test error", "path": ["importantEmail"]}]}
    Received:
      {"errors": [[GraphQLError: test error]]}
    Difference:
    - Expected
    + Received

    @@ -1,11 +1,11 @@
      Object {
        "errors": Array [
          Object {
            "locations": Array [
              Object {
    -           "column": 13,
    +           "column": 11,
                "line": 3,
              },
            ],
            "message": "test error",
            "path": Array [

      545 |       });
      546 | 
    > 547 |       expect(result).toMatchObject({
          |                      ^
      548 |         errors: [
      549 |           {
      550 |             message: "test error",

      at testReportsError (src/__tests__/subscription.test.ts:547:22)
      at Object.<anonymous> (src/__tests__/subscription.test.ts:516:5)
```
      
It seems like `column` is slightly off. Is this a problem?
      
The relevant code is at https://github.com/zalando-incubator/graphql-jit/blob/7937b3bee4164e26f902cbea5aff12d6ad48af5b/src/execution.ts#L1762

src/__tests__/subscription.test.ts
```diff
@@ -0,0 +1,1218 @@
...
+    await testReportsError(subscriptionRejectingErrorSchema);
+
+    async function testReportsError(schema: GraphQLSchema) {
+      // Promise<AsyncIterable<ExecutionResult> | ExecutionResult>
+      const result = await subscribe({
+        schema,
+        document: parse(`
+        subscription {
+          importantEmail
+        }
```
ruiaraujo, line 543: I would move this block two spaces forward if you want to keep the test values similar to graphql-js.

hoangvvo: That seems to do the trick. Thanks!

src/execution.ts
```diff
@@ -232,7 +240,17 @@ export function compileQuery(
...
+      context,
+      type,
+      context.operation.selectionSet,
+      Object.create(null),
+      Object.create(null)
+    );
+
+    const functionBody = compileOperation(context, type, fieldMap);
+
     const compiledQuery: InternalCompiledQuery = {
```
ruiaraujo, line 254: This is not supposed to be there in case of subscriptions, right?

hoangvvo: `compiledQuery.query()` is still be there. It will be used in `mapSourceToResponse` function similar to the way `graphql-js` uses `execute`. (with `rootValue` being the source)

https://github.com/graphql/graphql-js/blob/main/src/subscription/subscribe.js#L143

hoangvvo:
Is there anything I can do to push this forward? Thanks.

ruiaraujo:
Can you fix the build, we can get it in and then iterate on the API if needed.

hoangvvo:
I believe all of the changes I made have been covered by tests. I am not sure why coverage is dropped below 94%.

ruiaraujo:
:+1:

ruiaraujo:
@boopathi please take a look, I intend to merge this as is and iterate on the API if needed.

boopathi:
I was trying it locally with GraphQL-Tools' `makeExecutableSchema` and it required me to specify the default resolver as the identity function `x => x`. Let me know if I'm using it wrong.

The source code for this test is available here in this gist - https://gist.github.com/boopathi/885f41720f258b50a2ba67bed76581af 

The resolver is highlighted in the gist. Is it possible to not specify this resolver explicitly?

src/execution.ts
```diff
@@ -1637,3 +1669,194 @@ function getParentArgIndexes(context: CompilationContext) {
...
+    // GraphQL `execute` function, with `payload` as the rootValue.
+    // This implements the "MapSourceToResponseEvent" algorithm described in
+    // the GraphQL specification. The `execute` function provides the
+    // "ExecuteSubscriptionEvent" algorithm, as it is nearly identical to the
+    // "ExecuteQuery" algorithm, for which `execute` is also used.
+    // We use our `query` function in place of `execute`
+    const mapSourceToResponse = (payload: any) =>
+      queryFn(payload, executionContext.context, executionContext.variables);
+
+    return mapAsyncIterator(
```
boopathi, line 1797: Probably an unimportant issue - the arguments to the `next()` fn in the async iterator is ignored in the implementation of the function `mapAsyncIterator` - 

https://github.com/graphql/graphql-js/blob/fe9ea6db90a12c88d6372f88b61ca68e7aa3659b/src/subscription/mapAsyncIterator.js#L42-L43

I was trying to pass back a value from the GraphQL subscription result to the async generator function - 

```ts
// implementation
async *subscribe() {
  let previousValue = initialValue ?? previousValueFromInput;
  for (;;) {
    previousValue = yield promise(previousValue);
  }
}
```

```ts
// usage
const sub = await subscribe(`subscription { value }`);
const firstValue = await sub.next(
  something, client, wants, to, send, during, iteration /* These arguments are ignored */
);
```

This is probably a GraphQL-JS bug, I'm just curious if there are no usages of this pattern. We do not have to resolve this comment in this PR.

hoangvvo:
> I was trying it locally with GraphQL-Tools' `makeExecutableSchema` and it required me to specify the default resolver as the identity function `x => x`. Let me know if I'm using it wrong.
> 
> The source code for this test is available here in this gist - https://gist.github.com/boopathi/885f41720f258b50a2ba67bed76581af
> 
> The resolver is highlighted in the gist. Is it possible to not specify this resolver explicitly?

@boopathi 

That is strange. I wonder if GraphQL-Tools is doing something differently but the current test does not require resolverFn to be an identity function and it still passes:

https://github.com/zalando-incubator/graphql-jit/blob/07aac0e474818a7958bf4a270a2d0b9ef90f908e/src/__tests__/subscription.test.ts#L141

In your gist, if I replace the graphql-tools schema with the one below, it passes:

```js
const schema = new GraphQLSchema({
  subscription: new GraphQLObjectType({
    name: "Subscription",
    fields: {
      importantEmail: {
        type: NodeType,
        // resolve(source) {
       //   return source;
       // },
        async *subscribe() {
          for (let i = 0; i < 3; i++) {
            yield {
              id: i.toString()
            };
          }
        },
      }
    }
  })
});
```

src/execution.ts
hoangvvo: That is indeed a valid pattern. Perhaps we can implement our own version of mapAsyncIterator?

hoangvvo:
Looks like `graphql-tools` is also broken too for the `subscribe` function from graphql-js:

https://gist.github.com/hoangvvo/fa2901c5707eed116144426563fd6009

hoangvvo:
Odd because in `subscription.test.ts`, there are cases where resolve function is not set

src/execution.ts
hoangvvo: Yup mapAsyncIterator was changed to named export:

https://github.com/APIs-guru/graphql-js/blob/28315cd3f51b5ac212284f8583fb481053fa31a0/src/subscription/mapAsyncIterator.js

hoangvvo:
I am closing this in favor of https://github.com/zalando-incubator/graphql-jit/pull/115.

PR #96, opened 2021-01-21
feat: add support for @skip and @include directives

### LINKED ISSUE #68: Support skip and include
I know there is already a thread open about this but its almost a year old.
We really want to use this lib but the lack of this feature is why we cannot use it.

Any plans on implementing it? I am taking a look at the code to see if I would be able to send a PR but it looks a bit complicated for someone with no context to help, but I can spend more time on this probably.

Anyway, for the creators of this lib, implementing it would be much quicker and since `@skip` and `@include` are so widely used, is there anyway we can move forward with this feature?

Thanks and congrats on the great lib

boopathi:
An alternative try at implementation of `@skip` and `@include` directives.

previous effort: #70 

+ fix #70 
+ fix #68 

This approach generates the code in `ast.ts` during the field collection process. The advantage from the previous approach is that during this phase, we are able to add the skip/include conditions to sub-fields by augmenting the condition at every step.

For example,

```graphql
{
  foo @skip(if: $skip1) {
    bar @skip(if: $skip2)
  }
  ... {
    foo @skip(if: $skip3) {
      bar
    }
  }
}
```

Here, including `foo` depends on `$skip1` and `$skip3`.
But, including `bar` is trickier. It depends on `$skip1`, `$skip3` and `$skip2`.

TODO:

+ [x] Fix todo comments

### DIFF
--- a/src/__tests__/directives.test.ts
+++ b/src/__tests__/directives.test.ts
@@ -31,14 +31,8 @@ function executeTestQuery(query: string, variables = {}, schema = testSchema) {
   const ast = parse(query);
   const compiled: any = compileQuery(schema, ast, "", { debug: true } as any);
   if (!isCompiledQuery(compiled)) {
-    console.error(compiled);
-    throw new Error("compilation failed");
+    return compiled;
   }
-  // console.log(
-  //   require("prettier").format(
-  //     compiled.__DO_NOT_USE_THIS_OR_YOU_WILL_BE_FIRED_compilation
-  //   )
-  // );
   return compiled.query(data, undefined, variables);
 }
 
@@ -475,7 +469,7 @@ describe("Execute: handles directives", () => {
 
     describe("nested fragments", () => {
       const query = `
-        query ($skip1: Boolean!, $skip2: Boolean!, $include1: Boolean!, $include2: Boolean) {
+        query ($skip1: Boolean!, $skip2: Boolean!, $include1: Boolean!, $include2: Boolean!) {
           ...x @skip(if: $skip1)
           ... @include(if: $include1) {
             foo {
@@ -537,7 +531,7 @@ describe("Execute: handles directives", () => {
 
     describe("nested - non top-level fields", () => {
       const query = `
-        query ($skip1: Boolean!, $skip2: Boolean!, $include1: Boolean!, $include2: Boolean) {
+        query ($skip1: Boolean!, $skip2: Boolean!, $include1: Boolean!, $include2: Boolean!) {
           foo {
             ...aFragment @skip(if: $skip1)
             ... @skip(if: $skip2) {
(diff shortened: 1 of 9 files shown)


src/ast.ts
```diff
@@ -107,54 +183,234 @@ export function collectFields(
...
+  /**
+   * Recursively traverse through sub-selection and combine `shouldInclude`s
+   * from parent and current ones.
+   */
+  function handle(selection: SelectionNode) {
+    switch (selection.kind) {
+      case Kind.FIELD: {
+        const jitFieldNode: JitFieldNode = selection;
+        jitFieldNode.__internalShouldInclude = joinShouldIncludeCompilations(
+          rootFieldNode.__internalShouldInclude ?? "",
```
boopathi, line 248: this should be parentFieldNode

src/execution.ts
```diff
@@ -764,7 +767,13 @@ function compileObjectType(
       }" but got: $\{${GLOBAL_INSPECT_NAME}(${originPaths.join(".")})}.\``
     )}), null) :`;
   }
-  body += "{";
+
+  /**
+   * iife is used to optionally attach different fields
+   * based on skip/include directives.
+   */
+  const iifeBody = genFn();
```
ruiaraujo, line 775: I think we should go for a different approach where we set it to undefined and avoid iife. The benchmarks show a regression even though they do not use skip/include.

src/__tests__/directives.test.ts
```diff
@@ -2,40 +2,44 @@
...
-          return "a";
-        }
-      },
-      b: {
-        type: GraphQLString,
-        resolve() {
-          return "b";
-        }
-      }
+const testSchema = makeExecutableSchema({
```
ruiaraujo, line 10: I would a test confirming the resolvers are not being called.

src/ast.ts
ruiaraujo: The main question is why a test is not broken. :wink:

src/execution.ts
```diff
@@ -119,6 +121,7 @@ const GRAPHQL_ERROR = "__context.GraphQLError";
 const GLOBAL_RESOLVE = "__context.resolve";
 const GLOBAL_PARENT_NAME = "__parent";
 const LOCAL_JS_FIELD_NAME_PREFIX = "__field";
+const LOCAL_OBJECT_VAR_NAME = "__object";
```
ruiaraujo, line 124: If we move to the undefined approach we will be able to remove this.

src/execution.ts
boopathi: done 👍

src/__tests__/directives.test.ts
boopathi: done 👍

src/execution.ts
boopathi: done 👍

src/__tests__/directives.test.ts
```diff
@@ -313,4 +312,441 @@ describe("Execute: handles directives", () => {
...
+      const result = executeTestQuery(
+        query,
+        { skip: true },
+        getSchema(mockResolver)
+      );
+      expect(result.data.foo).toBeUndefined();
+      expect(mockResolver).not.toHaveBeenCalled();
+    });
+
+    test("resolver should not be called if not skipped", async () => {
```
ruiaraujo, line 741:
```suggestion
    test("resolver should be called if not skipped", async () => {
```

ruiaraujo (commented):
I think it is looking really good.

The only impact now for queries that do not use skip/include is that there will be a bunch of ternary everywhere with the condition hardcoded to true.

I think it might be worth taking the extra step to clean this up.

boopathi:
👍

ruiaraujo:
According to #98, the increase is around 2%. That is not enough to delay merging this in.

ruiaraujo:
:+1:

calmonr:
Nice job @boopathi. Is there an ETA for this to be released @ruiaraujo? Thank you guys.

ruiaraujo:
@calmonr The PR is already up so it should happen tomorrow.

PR #98, opened 2021-01-26
Report size of compiled function in benchmark

ruiaraujo:
We are discussing in #96 whether it is worth improving code size of part of the feature development.

Adding the reporting to the benchmark to provide more data.

PR #101, opened 2021-01-27
compact generated code to calculate size

boopathi:
this is a patch request to #98 to compute the size of the function body after compact print - remove unnecessary whitespace.


ruiaraujo:
I kinda disagree because what we produce is what gets stored in memory. This captures a different number that is not relevant. Whitespace does matter since it takes space currently.

boopathi:
Once you use the string inside `new Function`, the whitespace does not matter. The string gets GCed, isn't it? We do not save the generated code in memory. Only when `debug` is true, we even store that string in the compiled object.

ruiaraujo:
It stays in memory. You can do function.toString() and it will have the original. If it was GCed, I would not care about code size.

boopathi:
🤦‍♂️. okay. Then we are better off optimizing the generate code.

PR #113, opened 2021-05-05
fix __internalShouldInclude condition when a fragment is spread on si…

### LINKED ISSUE #111: Empty object when use fragment inside @skip and @include
Hello,
Im trying to integrate graphql-jit into an existing Apollo Server Service, we detected some queries sended by clients that are not working correctly when we change to graphql-jit, i tracked it down to this edge case (ill use very simplified queries):
Steps to reproduce
Start example/blog-apollo-server and make a this query:

```
query TEST(
    $includeOnly: Boolean!
){
    post(id: "post:1") {
        title
        Author1: author @skip(if: $includeOnly) {
            ...authorFragment
        }
        Author2: author @include(if: $includeOnly) {
            ...authorFragment
        }
    }
}
fragment authorFragment on User {
    name
}
```
Along with this variables:
```
{
    "includeOnly": true
}
```

and you will get an empty Author object: 
```
{
  "data": {
    "post": {
      "title": "Introduction to GraphQL!",
      "Author1": {}
    }
  }
}
```
it seems to be working if i remove the fragment and use props directly

fjmoreno:
fix #111 

When a fragment is used as child on multiple skip/include combinations the "__internalShouldInclude" computed condition is conflicting itself.

For example, consider this query
```
query TEST(
    $includeOnly: Boolean!
){
    post(id: "post:1") {
        title
        Author1: author @skip(if: $includeOnly) {
            ...authorFragment
        }
        Author2: author @include(if: $includeOnly) {
            ...authorFragment
        }
    }
}
fragment authorFragment on User {
    name
}
```
Along with this variables:
```
{
    "includeOnly": true
}
```

On this case, on Name field inside authorFragment the computed condition  on __internalShouldInclude  is 
```
(__context.variables["includeOnly"] === false) && (__context.variables["includeOnly"] === false) && (__context.variables["includeOnly"] === true) && (__context.variables["includeOnly"] === false) && (__context.variables["includeOnly"] === true) && (__context.variables["includeOnly"] === false)
```
Obviously this condition is always false and the field is never included, the approach on this PR is to not concat __internalShouldInclude conditions with the parent when coming from a fragment spread taking into account the current path the fragment field and not all paths leading to it.

### DIFF
--- a/src/ast.ts
+++ b/src/ast.ts
@@ -239,14 +239,16 @@ function augmentFieldNodeTree(
    * Recursively traverse through sub-selection and combine `shouldInclude`s
    * from parent and current ones.
    */
-  function handle(parentFieldNode: JitFieldNode, selection: SelectionNode) {
+  function handle(parentFieldNode: JitFieldNode, selection: SelectionNode, comesFromFragmentSpread: boolean = false) {
     switch (selection.kind) {
       case Kind.FIELD: {
         const jitFieldNode: JitFieldNode = selection;
-        jitFieldNode.__internalShouldInclude = joinShouldIncludeCompilations(
-          parentFieldNode.__internalShouldInclude ?? "",
-          jitFieldNode.__internalShouldInclude ?? ""
-        );
+        if (!comesFromFragmentSpread) {
+          jitFieldNode.__internalShouldInclude = joinShouldIncludeCompilations(
+            parentFieldNode.__internalShouldInclude ?? "",
+            jitFieldNode.__internalShouldInclude ?? ""
+          );
+        }
         // go further down the query tree
         for (const selection of jitFieldNode.selectionSet?.selections ?? []) {
           handle(jitFieldNode, selection);
@@ -262,13 +264,12 @@ function augmentFieldNodeTree(
       case Kind.FRAGMENT_SPREAD: {
         const fragment = compilationContext.fragments[selection.name.value];
         for (const subSelection of fragment.selectionSet.selections) {
-          handle(parentFieldNode, subSelection);
+          handle(parentFieldNode, subSelection, true);
         }
       }
     }
   }
 }
-
 /**
  * Joins a list of shouldInclude compiled code into a single logical
  * statement.


boopathi:
Thanks for the PR. The change looks good. Can you also update the tests to include the new case mentioned in the issue/pr-description that was failing in the main branch?

fjmoreno:
Sure, i updated it

src/ast.ts
```diff
@@ -262,13 +264,12 @@ function augmentFieldNodeTree(
       case Kind.FRAGMENT_SPREAD: {
         const fragment = compilationContext.fragments[selection.name.value];
         for (const subSelection of fragment.selectionSet.selections) {
-          handle(parentFieldNode, subSelection);
+          handle(parentFieldNode, subSelection, true);
```
boopathi, line 267: is this change not applicable for InlineFragments?

fjmoreno: No, in fact our workaround until now was rewrite queries to use inline fragments, following same example if you make next query is working as expected in main branch:
```
query TEST(
    $includeOnly: Boolean!
){
    post(id: "post:1") {
        title
        Author1: author @skip(if: $includeOnly) {
            ...on User {
              name
          }
        }
        Author2: author @include(if: $includeOnly) {
            ...on User {
                name
            }
        }
    }
}
```

And for the field name on first inline fragment you get this correct computed condition:
```
(__context.variables["includeOnly"] === false) && (__context.variables["includeOnly"] === false) 
```

And second gets also the correct one:
```
(__context.variables["includeOnly"] === true) && (__context.variables["includeOnly"] === true) 
```

ruiaraujo: We should add tests to answer this question.

fjmoreno: i update the tests...

boopathi (approved):

ruiaraujo:
@fjmoreno can you fix the build? We can then merge it.

ruiaraujo:
:+1:

boopathi:
👍

ruiaraujo:
:+1:

PR #115, opened 2021-05-23
Implement GraphQL subscription

### LINKED ISSUE #74: Support createSourceEventStream?

hoangvvo:
This PR builds upon the previous one https://github.com/zalando-incubator/graphql-jit/pull/94.

I refactor the implementation and update with recent changes in `graphql-js`. 

The test has also been updated to include the latest test cases from `graphql-js` using https://github.com/graphql/graphql-js/commit/1bd65a3efe5d4c7dad9c7d1e018610054a962e93 (3 days ago). For the diff of our test cases and theirs (mainly because we don't support root resolver), see https://github.com/zalando-incubator/graphql-jit/pull/115/commits/6c8cfc621d4c09abf8db13b02406b66205ac4936.

Close #74

### DIFF
--- a/src/__tests__/subscription.test.ts
+++ b/src/__tests__/subscription.test.ts
@@ -1,5 +1,6 @@
 /**
  * Based on https://github.com/graphql/graphql-js/blob/main/src/subscription/__tests__/subscribe-test.js
+ * (commit 39b69da863fca34f0e921bb9ac6a1b99797e17cf)
  * This test suite includes certain deviations from the original:
  * graphql-jit does not support the root resolver pattern that this test uses
  * so the part must be rewritten to include that root resolver in `subscribe` of
@@ -17,7 +18,7 @@ import {
   parse,
   SubscriptionArgs
 } from "graphql";
-import { compileQuery, isCompiledQuery } from "../execution";
+import { compileQuery, isAsyncIterable, isCompiledQuery } from "../execution";
 
 type Email = {
   from: string,
@@ -51,34 +52,12 @@ const InboxType = new GraphQLObjectType({
   },
 });
 
-const QueryType = new GraphQLObjectType({
-  name: "Query",
-  fields: {
-    inbox: { type: InboxType },
-  },
-});
-
 const EmailEventType = new GraphQLObjectType({
   name: "EmailEvent",
   fields: {
     email: { type: EmailType },
     inbox: { type: InboxType },
   },
-});
-
-const emailSchema = new GraphQLSchema({
-  query: QueryType,
-  subscription: new GraphQLObjectType({
-    name: "Subscription",
-    fields: {
-      importantEmail: {
-        type: EmailEventType,
-        args: {
-          priority: { type: GraphQLInt },
-        },
-      },
-    },
-  }),
 });
 
 async function subscribe({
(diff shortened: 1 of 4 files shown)


src/__tests__/subscription.test.ts
```diff
@@ -0,0 +1,1057 @@
...
+
+    // If we receive variables that cannot be coerced correctly, subscribe() will
+    // resolve to an ExecutionResult that contains an informative error description.
+    const result = await subscribe({ schema, document, variableValues });
+    expect(result).toEqual({
+      errors: [
+        {
+          message:
+            // DIFF: 'Variable "$arg" got invalid value "meow"; Int cannot represent non-integer value: "meow"',
+            'Variable "$arg" got invalid value "meow"; Expected type Int; Int cannot represent non-integer value: "meow"',
```
hoangvvo, line 488: I think, if it is not too breaking, we should change this message to stay up to date with `graphql-js`. Wdyt?

hoangvvo:
@boopathi I have confirmed that your test case (https://gist.github.com/boopathi/885f41720f258b50a2ba67bed76581af) now works. We are covering it in this case: https://github.com/zalando-incubator/graphql-jit/blob/27c197cd2bf5c82e6a1e0999e4373b21af99499a/src/__tests__/subscription.test.ts#L203-L220

hoangvvo:
@ruiaraujo Could you please approve the CI run? Thanks!

ardatan:
Any updates for this PR?

hoangvvo:
> Any updates for this PR?
On my side, all tests are currently passing, so I'm just waiting for a review actually.

boopathi:
@hoangvvo 

> @boopathi I have confirmed that your test case (https://gist.github.com/boopathi/885f41720f258b50a2ba67bed76581af) now works

I'm trying it out **without** the resolve function mentioned in my test case. It still gives me an error. Looks like when using makeExecutableSchema, a resolve function that does `(x) => x` is necessary for subscriptions. I don't understand this limitation.

hoangvvo:
> @hoangvvo
> 
> > @boopathi I have confirmed that your test case (https://gist.github.com/boopathi/885f41720f258b50a2ba67bed76581af) now works
> 
> I'm trying it out **without** the resolve function mentioned in my test case. It still gives me an error. Looks like when using makeExecutableSchema, a resolve function that does `(x) => x` is necessary for subscriptions. I don't understand this limitation.

Yeah, as I mentioned in the other PR, that pattern is not supported by `graphql-js` either. I think the function can only either be non-async generator function, or a promise that resolved into a async iterator.

boopathi (approved):
LGTM 🎉

boopathi:
👍

src/__tests__/subscription.test.ts
oporkka: Yes, we can change it.

src/execution.ts
```diff
@@ -1637,3 +1668,214 @@ function getParentArgIndexes(context: CompilationContext) {
...
+    }
+
+    try {
+      return { value: await callback(result.value), done: false };
+    } catch (error) {
+      if (typeof iterator.return === "function") {
+        try {
+          await iterator.return();
+        } catch (e) {
+          /* ignore error */
```
oporkka, line 1856: Would it be possible to extend the comment here to explicitly state why we ignore the error?

oporkka:
👍

PR #120, opened 2021-07-06
fix empty [RangeError: invalid String length] or OOM

### LINKED ISSUE #117: [RangeError: invalid String length]
Hello,

Im getting an error when use skip/inlude in a deep node, this error comes from this function:
https://github.com/zalando-incubator/graphql-jit/blob/15ec15fb61e0f0998725147ebb1f75c00e829d22/src/ast.ts#L301-L331

More specifically from at .join on line 330, the reason of that is that this function deduplicate multiples trues but it fails to deduplicate of the conditions comes from a variables, repeating multiple times same condition on the __internalShouldInclude, 

for example, if you have a nested skip/include with a var:
```javascript
....
@skip(if: $includeOnly)
....
```

you can get this __internalShouldInclude:
```javascript
(__context.variables["includeOnly"] === false) && (__context.variables["includeOnly"] === false) &&  (__context.variables["includeOnly"] === false) && (__context.variables["includeOnly"] === false) && (__context.variables["includeOnly"] === false) && (__context.variables["includeOnly"] === false) &&  (__context.variables["includeOnly"] === false) && (__context.variables["includeOnly"] === false)
.......................................
```

Must deduplicate it to ensure not exceed v8 max string length

fjmoreno:
fix https://github.com/zalando-incubator/graphql-jit/issues/117

It prevents to duplicate shouldInclude conditions when using a fragments

### DIFF
--- a/src/__tests__/directives.test.ts
+++ b/src/__tests__/directives.test.ts
@@ -349,8 +349,8 @@ describe("Execute: handles directives", () => {
       }
     });
     test("it does not throw [RangeError: invalid String length] or OOM", async () => {
-        const schema = makeExecutableSchema({
-          typeDefs: `
+      const schema = makeExecutableSchema({
+        typeDefs: `
         type Query {
           firstLevel: FirstLevel
         }
@@ -368,32 +368,32 @@ describe("Execute: handles directives", () => {
           d: String
         }
       `,
-          resolvers: {
-            Query: {
-              firstLevel: () => ({ b: 42 })
-            },
-            FirstLevel: {
-              secondLevel() {
-                return {};
-              }
-            },
-            SecondLevel: {
-              thirdLevel() {
-                return {};
-              }
-            },
-            ThirdLevel: {
-              fourthLevel() {
-                return {
-                  c: "ccc",
-                  d: "ddd"
-                };
-              }
+        resolvers: {
+          Query: {
+            firstLevel: () => ({ b: 42 })
+          },
+          FirstLevel: {
+            secondLevel() {
+              return {};
+            }
+          },
+          SecondLevel: {
+            thirdLevel() {
+              return {};
+            }
+          },
+          ThirdLevel: {
+            fourthLevel() {
+              return {
+                c: "ccc",
+                d: "ddd"
+              };
             }
           }
-        });
+        }
+      });
 
-        let fragmentSpreadQuery = `query (
+      let fragmentSpreadQuery = `query (
             $includeVar: Boolean!,
             $includeVar2: Boolean!,
             $includeVar3: Boolean!,
@@ -403,8 +403,8 @@ describe("Execute: handles directives", () => {
             ) {
             `;
 
-        for (let i = 0; i < 500; i++) {
-          fragmentSpreadQuery += `
+      for (let i = 0; i < 500; i++) {
+        fragmentSpreadQuery += `
          iteration${i}:firstLevel @skip(if: $skipVar) @include(if: $includeVar) {
             secondLevel @include(if: $includeVar) @skip(if: $skipVar){
               thirdLevel @include(if: $includeVar2) @skip(if: $skipVar){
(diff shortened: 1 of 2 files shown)


src/__tests__/directives.test.ts
```diff
@@ -348,6 +348,123 @@ describe("Execute: handles directives", () => {
         }
       }
     });
+    test("it does not throw [RangeError: invalid String length] or OOM", async () => {
```
oporkka, line 351: The test case is testing that include in nested fragments work in general so moving the specific bug case to a line comment instead, as context.
```suggestion
    // Specifically for `[RangeError: invalid String length]` https://github.com/zalando-incubator/graphql-jit/issues/117
    test("it supports use in nested fragments", async () => {
```

src/ast.ts
```diff
@@ -299,34 +299,11 @@ function augmentFieldNodeTree(
  * @param compilations
  */
 function joinShouldIncludeCompilations(...compilations: string[]) {
-  // remove empty strings
```
oporkka, line 302: Why removing this line?

src/ast.ts
```diff
@@ -299,34 +299,11 @@ function augmentFieldNodeTree(
...
-  /**
-   * If we have filtered out everything in the list, it means, we
-   * called this with a join(true, true, etc...) - returning empty string
-   * is wrong, so we return a single true
-   */
-  if (filteredCompilations.length < 1) {
-    return "true";
-  }
-
+  filteredCompilations = Array.from(new Set(filteredCompilations));
```
oporkka, line 306: Would there be any helpful line comment to add for the future reader, why `Array.from(new Set(..))` is needed?

src/ast.ts
```diff
@@ -302,31 +302,12 @@ function joinShouldIncludeCompilations(...compilations: string[]) {
...
-  // So, the more nested the query is, the more of unnecessary `&& true`
-  // we get.
-  //
-  // Failing to do this results in [RangeError: invalid array length]
-  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Invalid_array_length
-  filteredCompilations = filteredCompilations.filter(
-    it => it.trim() !== "true"
+  // Split conditions by && and flatten it
+  filteredCompilations = ([] as string[]).concat(
+    ...filteredCompilations.map(e => e.split(" && "))
```
boopathi, line 307: let us disregard empty spaces, if any, in this deduplication.

```suggestion
    ...filteredCompilations.map(e => e.split(" && ").map(it => it.trim()))
```

src/ast.ts
```diff
@@ -302,31 +302,12 @@ function joinShouldIncludeCompilations(...compilations: string[]) {
...
-  // remove "true" since we are joining with '&&' as `true && X` = `X`
-  // This prevents an explosion of `&& true` which could break
-  // V8's internal size limit for string.
-  //
-  // Note: the `true` appears if a field does not have a skip/include directive
-  // So, the more nested the query is, the more of unnecessary `&& true`
-  // we get.
-  //
-  // Failing to do this results in [RangeError: invalid array length]
-  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Invalid_array_length
```
boopathi, line 316: It's still valuable to retain this comment and update it. The FIXME is no longer required. But the explanation of why deduplication exists would help.

src/__tests__/directives.test.ts
oporkka: I would still like to have the tested edge case in the test name, instead of the impact of the bug that we had in the implementation. It will be more helpful to a future developer who sees this test failing.
```suggestion
    test("it supports use in nested fragments", async () => {
```

src/ast.ts
oporkka: Should we have a test for this?

boopathi (approved):
Thank you for the fix.

boopathi:
👍

oporkka:
:+1:

MarusDod:
still ocurring this error in 8.0.4

oporkka:
> still ocurring this error in 8.0.4

@MarusDod would you be able to provide a test case for the problem, similar to as in this PR?

PR #139, opened 2021-10-04
Add TypedDocumentNode support

ardatan:
Thank you for this great library!
This PR adds `TypedDocumentNode` support to `graphql-jit`. For example, generated `TypedDocumentNode` queries will have type safe signature with `graphql-jit`.
It also enables to use generics in `CompiledQuery` to have better type safety.
https://github.com/dotansimha/graphql-typed-document-node

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,6 +75,7 @@
     "typescript": "^3.9.5"
   },
   "dependencies": {
+    "@graphql-typed-document-node/core": "3.1.0",
     "fast-json-stringify": "^1.13.0",
     "generate-function": "^2.3.1",
     "json-schema": "^0.2.3",
--- a/src/execution.ts
+++ b/src/execution.ts
@@ -1,3 +1,4 @@
+import { TypedDocumentNode } from "@graphql-typed-document-node/core";
 import fastJson from "fast-json-stringify";
 import genFn from "generate-function";
 import {
@@ -61,7 +62,6 @@ import {
   compileVariableParsing,
   failToParseVariables
 } from "./variables";
-import { TypedDocumentNode } from "@graphql-typed-document-node/core";
 
 const inspect = createInspect();
 
@@ -163,7 +163,10 @@ interface DeferredField {
   args: Arguments;
 }
 
-export interface CompiledQuery<TResult = { [key: string]: any }, TVariables = { [key: string]: any }> {
+export interface CompiledQuery<
+  TResult = { [key: string]: any },
+  TVariables = { [key: string]: any }
+> {
   operationName?: string;
   query: (
     root: any,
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -75,6 +75,7 @@
     "typescript": "^3.9.5"
   },
   "dependencies": {
+    "@graphql-typed-document-node/core": "3.1.0",
```
hoangvvo, line 78: Is this for TS annotation only? In that case I think it is better to be in `devDependencies`

ardatan: `TypedDocumentNode` is used in the exported types and functions such as `compileQuery` and `CompiledQuery`. If we move this to `devDependencies`, those types and function signatures will break.

ardatan:
Is there anything I can do to get this merged?

dotansimha:
> > Is there anything I can do to get this merged?
> 
> We have to get two Zalando employees to review this PR. As the reviewers might not be familiar with `TypedDocumentNode`, my proposal would be to extend the PR description to extend why adding support for `TypedDocumentNode` is needed. I see it's about type safety, but how would the experience of users of GraphQL-jit improve? Do you have any code examples? Do the current users need to change their code somehow if this gets merged and they update GraphQL-jit?

Sure. 

GraphQL-Codegen generates a ready-to-use `TypedDocumentNode` (it's basically a `DocumentNode` with the response and variables types burned into the type). 
When a user passes this `TypedDocumentNode` to GraphQL-JIT's execute function, it gets type-safety our of the box:

```
const executionResult = await compiledQuery.query(root, context, variables);
// Now, executionResult is fully typed automatically, and variables is also checked. 
```

This way, developers don't need to specify types manually at all. `TypedDocumentNode` is now adopted by `graphql-js`, `apollo-client`, `urql` and many other popular libraries.

ruiaraujo:
@ardatan please run the formatter. This PR lgtm!

ardatan:
I think it is good now! @ruiaraujo

ardatan:
I think it is good now! @ruiaraujo

ruiaraujo:
:+1:

ruiaraujo:
@ardatan i believe you are pushing code to the wrong branch.

ardatan:
@ruiaraujo Yes I realized that :) I cherrypicked those commits to the other PR

ardatan:
Ok I fixed lint/format issue because of the new import.

ruiaraujo:
:+1:

ardatan:
I think it is ready to merge right? @ruiaraujo

ruiaraujo:
Once a colleague from zalando approves, it will be merged.

ardatan:
@boopathi Could you review this?

PR #140, opened 2021-10-27
GraphQL v16 compatibility

ardatan:
GraphQL v16 has been released recently so this PR does the following changes for the compatibility;

- [x] Type resolvers no longer returns type itself but the name of the type because this no longer supported in v16
- [x] Update TS version because v16 doesn't support existing version
- [x] Add tests for both v15 and v16
- [x] Use the custom implementation of `collectFields` in `json.ts` because it has been changed in v16
- [x] Fix typing issues
- [x] Use isSubType instead of isPossibleType because it has been removed in v16
- [x] Replace `toEqual` with `toMatchObject` to compare results because `GraphQLError` is no longer serialized like JSO
- [x] Remove Node 10 from CI because v16 doesn't support it and it has been no longer maintained.
- [x] Update TS version to support latest changes and v16
- [x] Migrate to ESlint and update Prettier to support new TypeScript ESLint plugin

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,9 +78,6 @@
     "prettier": "^2.4.1",
     "ts-jest": "^24.0.2",
     "ts-node": "^8.0.3",
-    "tslint": "^5.15.0",
-    "tslint-config-prettier": "^1.18.0",
-    "tslint-plugin-prettier": "^2.0.1",
     "typescript": "^4.4.4"
   },
   "dependencies": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -1068,7 +1068,7 @@ buffer-from@1.x, buffer-from@^1.0.0:
   resolved "https://registry.yarnpkg.com/buffer-from/-/buffer-from-1.1.1.tgz#32713bc028f75c02fdb710d7c7bcec1f2c6070ef"
   integrity sha512-MQcXEUbCKtEo7bhqEs6560Hyd4XaovZlO/k9V3hjVUF/zwW7KBVdSK4gIt/bzwS9MbR5qob+F5jusZsb0YQK2A==
 
-builtin-modules@^1.0.0, builtin-modules@^1.1.1:
+builtin-modules@^1.0.0:
   version "1.1.1"
   resolved "https://registry.yarnpkg.com/builtin-modules/-/builtin-modules-1.1.1.tgz#270f076c5a72c02f5b65a47df94c5fe3a278892f"
   integrity sha1-Jw8HbFpywC9bZaR9+Uxf46J4iS8=
@@ -1170,7 +1170,7 @@ chalk@^2.0.0, chalk@^2.0.1:
     escape-string-regexp "^1.0.5"
     supports-color "^5.3.0"
 
-chalk@^2.3.0, chalk@^2.3.1, chalk@^2.4.1, chalk@^2.4.2, chalk@~2.4.1:
+chalk@^2.3.1, chalk@^2.4.1, chalk@^2.4.2, chalk@~2.4.1:
   version "2.4.2"
   resolved "https://registry.yarnpkg.com/chalk/-/chalk-2.4.2.tgz#cd42541677a54333cf541a49108c1432b44c9424"
   integrity sha512-Mti+f9lpJNcwF4tWV8/OrTTtF1gZi+f8FqlyAdouralcFWFQWF2+NgCHShjkCb+IFBLq9buZwE1xckQU4peSuQ==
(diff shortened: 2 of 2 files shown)


ardatan:
Anything needed for this PR?

ardatan:
Sorry it is my bad :( Could you run it again? @boopathi

ardatan:
I hate to ask but I need another :) Is it possible to make this automatic only for this PR ? @boopathi

ardatan:
I removed Node 10 because v16 doesn't support it and it has been no longer maintained.

ruiaraujo:
@ardatan switch from node 15 to node 16 on the build.

dotansimha:
@ruiaraujo  @boopathi @topicus can you please review this PR? seems like it's breaking on other tools as well (`envelop` / `graphql-mesh`) and blocks the ability to upgrade.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
ruiaraujo, line 1: Please remove this file.

ardatan: Done!

ardatan:
@ruiaraujo Done!

ardatan:
@ruiaraujo I also ran prettier and ts-lint to fix format/lint issues.

src/execution.ts
```diff
@@ -1493,7 +1489,7 @@ function defaultResolveTypeFn(
           `Promises are not supported for resolving type of ${value}`
         );
       } else if (isTypeOfResult) {
-        return type;
+        return type.name;;
```
ruiaraujo: Can you justify this change here?

ardatan: In GraphQL v16, type resolvers cannot return type itself but only the name of the type.

src/__tests__/abstract.test.ts
```diff
@@ -95,7 +95,7 @@ describe("Execute: Handles execution of abstract types", () => {
 
     const result = graphql(schema, query);
 
-    expect(result).toEqual({
+    expect(result).toMatchObject({
```
ruiaraujo, line 98: I have read the description which mentions having to change toEqual to MatchObject due to differences in the GraphQL error object.

How does it related here with no errors?

ardatan: It was looking weird while some of tests still have `toEqual`.

ruiaraujo: I am ok with mixing since toEqual is a much stronger check than MatchObject.
I would rather only change the ones that were failing.

ruiaraujo:
@ardatan the build is still complaining about the formatter.

ardatan:
@ruiaraujo We need TS v4 for GraphQL v16, TS v4 isn't supported by TSLint and TSLint doesn't know about the following syntax;
```
} catch(e: any) { //Casting errors which necessary for TS
```
TSLint is deprecated and should be replaced with ESLint.

ruiaraujo:
@ardatan we also need to conver https://github.com/zalando-incubator/graphql-jit/blob/master/tslint.json to eslint.json

ardatan:
I think rules are fine now since TSLint rules don't exist as-is in ESLint

ruiaraujo:
@ardatan almost there, you're getting closer. :smile:

ardatan:
I hope this time it will pass @ruiaraujo  :)

ruiaraujo:
:+1:

ardatan:
I think you need to update repo settings for "Expected" workflows

ruiaraujo:
@ardatan we will do it when we are ready to merge with the second review, no worries. Thanks for the great contribuition.

addityasingh:
@ardatan can you fix the conflicts since we merged your other PR #139 ?

dotansimha (approved):

ardatan:
Done @addityasingh @ruiaraujo @boopathi

ruiaraujo:
:+1:

ardatan:
@ruiaraujo @boopathi @topicus @addityasingh When could this get merged and released? We really need this for our libraries in order to upgrade to GraphQL v16.

package.json
```diff
@@ -61,18 +61,27 @@
...
+     "graphql": "^16.0.0",
     "jest": "^24.7.1",
     "lint-staged": "^8.1.5",
-    "prettier": "^1.19.1",
+    "prettier": "^2.4.1",
     "ts-jest": "^24.0.2",
     "ts-node": "^8.0.3",
     "tslint": "^5.15.0",
     "tslint-config-prettier": "^1.18.0",
     "tslint-plugin-prettier": "^2.0.1",
```
oporkka, line 83: Will `tslint` dependencies still be needed even if we move to `eslint`?

src/error.ts
```diff
@@ -64,11 +60,6 @@ export function GraphQLError(
   UpstreamGraphQLError.prototype,
   {
     constructor: { value: GraphQLError },
-    name: { value: "GraphQLError" },
-    toString: {
```
oporkka, line 68: Which part of the update requires removing `toString`?

ardatan: `printError` calls `toString` in v16 so this causes an infinite loop.
You can see the relevant change in v16;
https://github.com/graphql/graphql-js/pull/3252

package.json
ardatan: Nice catch! Removed!

ardatan: My bad :) I forgot to push

ardatan: Could you take a look if it is good for you? @oporkka

ruiaraujo:
:+1:

oporkka:
:+1:

ardatan:
Thank you <3

Issue #142, opened 2021-11-05
Input type defaults not supported?


laurisvan:
Our stack is using a combination of apollo-server-express (to be removed), graphql-envelop, graphql-modules, and now recently we were adding graphql-jit.

While the stack works alright otherwise, it appears that out input type defaults do not get applied in mutations. A sample:

```
input CustomerInput {
  id: ID
  type: CustomerType = WORK_SITE
}
```

Before digging further, I'd like to get your confirmation whether input type defaults should be working, or is this something new? Could this be related to using ENUMs or something else? Is there a working sample/test/whatever that I could try and extend?

laurisvan:
Effectively this works when the type is set not nullable. I wonder if this is somehow related to https://github.com/zalando-incubator/graphql-jit/issues/91 ?

ruiaraujo:
I need to check the spec to understand the right behaviour. If you contribute a failing test, it will help. :)

laurisvan:
@ruiaraujo Will be glad to write one. Can you give a bit of hint on the structure of tests in this repo - where should such a test go? 

Could https://github.com/zalando-incubator/graphql-jit/blob/master/src/__tests__/variables.test.ts 
```
const TestInputObject = new GraphQLInputObjectType({
  name: "TestInputObject",
  fields: {
    a: { type: GraphQLString },
    b: { type: new GraphQLList(GraphQLString) },
    c: { type: new GraphQLNonNull(GraphQLString) },
    d: { type: TestComplexScalar },
    e: { type: new GraphQLList(GraphQLString) }
  }
});
```

be extended with an input type with a default?

laurisvan:
I think the corresponding spec is https://spec.graphql.org/June2018/#sec-Input-Objects, "Input Coercion":

```
If a variable is provided for an input object field, the runtime value of that variable must be used. If the runtime value is null and the field type is non‐null, a field error must be thrown. If no runtime value is provided, the variable definition’s default value should be used. If the variable definition does not provide a default value, the input object field definition’s default value should be used.
```

I'm not fully sure how graphql-jit implementation of null values works - My reading of the spec tells that if the input field is specifically given as null, it should be treated as null; but if missing, should be coerced to default.

KariHe:
I have created unit test to demonstrate the issue. Tests can be found from https://github.com/KariHe/graphql-jit/tree/input-defaults-tests (Should I open PR to get them in?)

When input is given directly in request defaults are working.
`query {
[...]

PR #143, opened 2021-11-10
Unit tests for input defaults

KariHe:
Added unit tests for testing that input default values are working. Currently default values are not applied for inputs that are coming from variables that is discussed in issue #142.

### DIFF
--- a/src/__tests__/variables.test.ts
+++ b/src/__tests__/variables.test.ts
@@ -75,7 +75,7 @@ const TestInputObjectWithDefaultValue = new GraphQLInputObjectType({
   name: "TestInputObjectWithDefaultValue",
   fields: {
     a: { type: GraphQLString },
-    b: { type: GraphQLString, defaultValue: "default" },
+    b: { type: GraphQLString, defaultValue: "default" }
   }
 });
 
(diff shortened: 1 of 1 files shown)


ruiaraujo:
@KariHe can you run the formatter please?

src/__tests__/variables.test.ts
```diff
@@ -1577,6 +1588,39 @@ describe("Execute: Handles inputs", () => {
...
+      `);
+
+      expect(result).toEqual({
+        data: {
+          fieldWithDefaultInputValue: '{ a: "test", b: "default" }'
+        }
+      });
+    });
+
+    test("allows inputs with default values to be ommited from variable", async () => {
```
ruiaraujo, line 1606:
```suggestion
    test.skip("allows inputs with default values to be ommited from variable", async () => {
```

ruiaraujo:
:+1:

oporkka:
:+1:

src/__tests__/variables.test.ts
```diff
@@ -1577,6 +1588,39 @@ describe("Execute: Handles inputs", () => {
     });
   });
 
+  describe("Handles default values", () => {
```
addityasingh, line 1591: How is this test different from `"Execute: Uses argument default values.when no argument provided"`?

KariHe: Test uses input type that have field with default value. `"Execute: Uses argument default values.when no argument provided"` uses default value on basic type input instead. So the basic case where this issue comes is like this:

```
input TestInput {
  str: String
  type: String = "default"
}
type Query {
  test(input: TestInput!): String
}
```

PR #145, opened 2021-11-12
Properly take defaults into account.

### LINKED ISSUE #142: Input type defaults not supported?

ruiaraujo:
closes #142

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "graphql-jit",
-  "version": "0.7.0",
+  "version": "0.7.1",
   "description": "GraphQL JIT Compiler to JS",
   "main": "dist/index.js",
   "types": "./dist/index.d.ts",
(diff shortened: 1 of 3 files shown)


ruiaraujo:
:+1:

boopathi:
👍

topicus:
👍

PR #152, opened 2021-12-03
Better error message in default type resolver

ardatan:
The error is so cryptic if type isn't resolved by default type resolver. So this PR improves that issue.


ruiaraujo:
👍

boopathi:
👍

PR #153, opened 2021-12-07
[RFC] compile createSourceEventStream

Dschoordsch:
Creating a source event stream from the compiled query is about 1.5 to 2 times faster. The use case for createSourceEventStream is running the event stream and resolvers in different processes. That means compiling the resolvers is not necessary and just costs performance, thus a separate function is exposed for just this use case.
It does not actually compile anything, but that's an implementation detail und thus `compileCreateSourceEventStream` name was chosen.

This needs some polishing and tests, but I would like to know if you're interested in this at all?

Relates-To: https://github.com/ParabolInc/parabol/issues/5468

### DIFF
--- a/src/__benchmarks__/benchmarks.ts
+++ b/src/__benchmarks__/benchmarks.ts
@@ -1,5 +1,6 @@
 import Benchmark from "benchmark";
 import {
+  createSourceEventStream,
   DocumentNode,
   execute,
   getIntrospectionQuery,
@@ -7,6 +8,7 @@ import {
   parse
 } from "graphql";
 import { compileQuery, isCompiledQuery, isPromise } from "../execution";
+import { benchmarkCreateSourceEventStream } from "./createSourceEventStream";
 import {
   query as fewResolversQuery,
   schema as fewResolversSchema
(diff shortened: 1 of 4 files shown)


Dschoordsch:
@markrzen could you take a look and let me know if this is something you would be willing to accept? If yes, I will polish it up.

markrzen:
Hey @Dschoordsch, I am not a maintainer of the library. You will want to ping the good folks on this list:
https://github.com/zalando-incubator/graphql-jit/blob/main/MAINTAINERS

ruiaraujo:
I wont have time before the weekend to check it.

Dschoordsch:
@ruiaraujo did you have chance to take a look whether or not a `compileCreateSourceEvent` function would be something you would accept?

Dschoordsch:
@boopathi could you take a look if that's something which you would be interested in?

oporkka:
@Dschoordsch We would like to learn if there are more needs for this in the GraphQL JIT user community

laurisvan:
> @Dschoordsch We would like to learn if there are more needs for this in the GraphQL JIT user community

Subscriptions are essential for our service, so at least we'd be interested in trying this out. I've checked createSourceEventStream in graphql before, and frankly speaking I'm not fully aware of how it works these days. Probably some kind of "for dummies" explanation would be great. :)

Dschoordsch:
Oh wow, I totally forgot and did not expect activity here. I think in most cases you can ignore `createSourceEventStream` if you're subscribing and resolving in the same process. For us however the websockets and subscriptions live in one process and the resolving happens in a different one and they're communicating via redis. That's why we need to split these 2 parts up. [Here](https://github.com/ParabolInc/parabol/blob/8e0502e3e588618eb936b271d8afdeeb931a4148/docs/serviceDiagram.md) you can get an idea of how it works. To be honest, it's been some time, so I would need to spend a bit of time to get back into the topic.
If the maintainers are interested, I would still be up to getting this in a production state.

Dschoordsch:
This is outdated.

PR #156, opened 2022-02-04
fix: update @graphql-typed-document-node/core for graphql@16 support

SimenB:
Not sure why the version is locked down, but 3.1.0 does not have `graphql@16` in its peer dependency range

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "typescript": "^4.4.4"
   },
   "dependencies": {
-    "@graphql-typed-document-node/core": "3.1.0",
+    "@graphql-typed-document-node/core": "3.1.1",
     "fast-json-stringify": "^1.13.0",
     "generate-function": "^2.3.1",
     "json-schema": "^0.4.0",
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "typescript": "^4.4.4"
   },
   "dependencies": {
-    "@graphql-typed-document-node/core": "3.1.1",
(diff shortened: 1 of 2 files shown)


mcollina:
This might explains I few of the problems I hit, good spot @SimenB!

package.json
```diff
@@ -81,7 +81,7 @@
     "typescript": "^4.4.4"
   },
   "dependencies": {
-    "@graphql-typed-document-node/core": "3.1.0",
```
oporkka, line 84: This was contributed in https://github.com/zalando-incubator/graphql-jit/pull/139

@ardatan do you recall any special reason why we would have locked the version to `3.1.0`?

oporkka:
@SimenB please run also `yarn install` to update the lockfile

oporkka:
Sorry for accidental close

package.json
SimenB: yeah, ideally it'd just be `^3.1.0` I guess 🙂

package.json
```diff
@@ -81,7 +81,7 @@
...
+    "@graphql-typed-document-node/core": "3.1.1",
```
oporkka, line 84:
```suggestion
    "@graphql-typed-document-node/core": "^3.1.1",
```

ruiaraujo:
👍

oporkka:
:+1:

SimenB:
Thanks! Could you release this as well? 😀

ruiaraujo:
@SimenB going out now, thanks for the contribuition.

PR #158, opened 2022-02-09
fix: parse Date scalar object arguments properly

### LINKED ISSUE #157: Non-primitive custom scalar argument gets coerced to a string
I tried adding `graphql-jit` to a project of mine and ran into some type errors with resolvers that have custom `Date` scalars as arguments. I made a failing test demonstrating the problem here: https://github.com/andrew0/graphql-jit/commit/abf1adde5c72769febb6514e7ba7940f19689a2f.

```ts
it("deserializes non-primitive scalars properly", async () => {
  const request = `
    {
      scalar(arg: "2022-01-01")
    }
  `;

  const schema = new GraphQLSchema({
    query: new GraphQLObjectType({
      name: "Query",
      fields: {
        scalar: {
          type: GraphQLString,
          args: {
            arg: {
              type: new GraphQLScalarType({
                name: "Date",
                serialize: (value: any) => value.toISOString().slice(0, 10),
                parseValue: (value: any) => new Date(value),
                parseLiteral: (ast) =>
                  ast.kind === Kind.STRING ? new Date(ast.value) : null
              })
            }
          },
          resolve: (_parent, { arg }) => Object.prototype.toString.call(arg)
        }
      }
    })
  });

  const result = await executeQuery(schema, parse(request));
  expect(result).toEqual({
    data: {
      scalar: "[object Date]"
    }
  });
});
```

With the regular `graphql-js` engine, this test passes, but with `graphql-jit`, `arg` gets converted to the string `2022-01-01T00:00:00.000Z` (i.e. the result of calling `toJSON()` on a `Date` object).

andrew0:
Fixes #157.

### DIFF
--- a/src/execution.ts
+++ b/src/execution.ts
@@ -1160,10 +1160,11 @@ function safeMap(
 }
 
 const MAGIC_MINUS_INFINITY =
-  "__MAGIC_MINUS_INFINITY__71d4310a-d4a3-4a05-b1fe-e60779d24998";
+  "__MAGIC_MINUS_INFINITY__71d4310a_d4a3_4a05_b1fe_e60779d24998";
 const MAGIC_PLUS_INFINITY =
-  "__MAGIC_PLUS_INFINITY__bb201c39-3333-4695-b4ad-7f1722e7aa7a";
-const MAGIC_NAN = "__MAGIC_NAN__57f286b9-4c20-487f-b409-79804ddcb4f8";
+  "__MAGIC_PLUS_INFINITY__bb201c39_3333_4695_b4ad_7f1722e7aa7a";
+const MAGIC_NAN = "__MAGIC_NAN__57f286b9_4c20_487f_b409_79804ddcb4f8";
+const MAGIC_DATE = "__MAGIC_DATE__33a9e76d_02e0_4128_8e92_3530ad3da74d";
 
 function specialValueReplacer(_: any, value: any) {
   if (Number.isNaN(value)) {
@@ -1183,9 +1184,9 @@ function specialValueReplacer(_: any, value: any) {
 
 function objectStringify(val: any): string {
   return JSON.stringify(val, specialValueReplacer)
-    .replace(`"${MAGIC_NAN}"`, "NaN")
-    .replace(`"${MAGIC_PLUS_INFINITY}"`, "Infinity")
-    .replace(`"${MAGIC_MINUS_INFINITY}"`, "-Infinity");
+    .replace(new RegExp(`"${MAGIC_NAN}"`, "g"), "NaN")
+    .replace(new RegExp(`"${MAGIC_PLUS_INFINITY}"`, "g"), "Infinity")
+    .replace(new RegExp(`"${MAGIC_MINUS_INFINITY}"`, "g"), "-Infinity");
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/execution.ts
```diff
@@ -1160,12 +1160,13 @@ function safeMap(
 }
 
 const MAGIC_MINUS_INFINITY =
-  "__MAGIC_MINUS_INFINITY__71d4310a-d4a3-4a05-b1fe-e60779d24998";
+  "__MAGIC_MINUS_INFINITY__71d4310a_d4a3_4a05_b1fe_e60779d24998";
 const MAGIC_PLUS_INFINITY =
-  "__MAGIC_PLUS_INFINITY__bb201c39-3333-4695-b4ad-7f1722e7aa7a";
-const MAGIC_NAN = "__MAGIC_NAN__57f286b9-4c20-487f-b409-79804ddcb4f8";
+  "__MAGIC_PLUS_INFINITY__bb201c39_3333_4695_b4ad_7f1722e7aa7a";
+const MAGIC_NAN = "__MAGIC_NAN__57f286b9_4c20_487f_b409_79804ddcb4f8";
```
andrew0, line 1166: I replaced the dashes with underscores, since the dashes need to be escaped in a regex sometimes.

src/execution.ts
```diff
@@ -1178,14 +1179,19 @@ function specialValueReplacer(_: any, value: any) {
...
 }
 
 function objectStringify(val: any): string {
   return JSON.stringify(val, specialValueReplacer)
-    .replace(`"${MAGIC_NAN}"`, "NaN")
-    .replace(`"${MAGIC_PLUS_INFINITY}"`, "Infinity")
-    .replace(`"${MAGIC_MINUS_INFINITY}"`, "-Infinity");
+    .replace(new RegExp(`"${MAGIC_NAN}"`, "g"), "NaN")
+    .replace(new RegExp(`"${MAGIC_PLUS_INFINITY}"`, "g"), "Infinity")
+    .replace(new RegExp(`"${MAGIC_MINUS_INFINITY}"`, "g"), "-Infinity")
```
andrew0, line 1193: I noticed that these were only replacing the first occurrence of the magic strings, so I changed them to use a global regex.

boopathi: We could use replaceAll instead of making it a RegExp.

boopathi:
Thanks for your PR! The PR looks good.

Though this solves the problem specifically for Date objects, I wonder if it can be better handled for any custom class. I’ll review the change more thoroughly this week.

src/execution.ts
andrew0: replaceAll wasn't added to Node until v15, so you'd lose support for maintenance LTS versions.

boopathi: Ah. Cool!

src/execution.ts
boopathi: Sounds good.

flashspys:
As Date is a very popular object, is it possible to merge this MR already and open a new issue for custom transformers?

oporkka:
:+1:

boopathi:
👍

boopathi:
We will explore more about custom classes in future. For now supporting Date looks important. Thank you!

flashspys:
It would be awesome, if you could create a new release with this change ❤️

tuan231195:
please create a new release with this change

oporkka:
> please create a new release with this change

@tuan231195 @flashspys sorry for the delay. There is now a release with this change.

PR #164, opened 2022-02-15
fix: update deprecated API from graphql-js

boopathi:
`getOperationRootType` is deprecated in graphql-js. The new API is `schema.getRootType(operation)`

Deprecation notice

https://github.com/graphql/graphql-js/blob/47bd8c8897c72d3efc17ecb1599a95cee6bac5e8/src/utilities/getOperationRootType.ts#L11-L15


boopathi:
This is not a supported API in 15. We have to wait till we upgrade to graphql-js v16,v17.

PR #177, opened 2022-08-25
make new function exposed from the enricher so lookaheads can check and evaluate skips / includes 

trkohler:
### Bug description
Even though fields inside the entities() field of a collection are @skip-ed as part of a query, they still make the calls to backends as if they are not skipped but just discards the data instead of returning it to the client.

### Expected behavior
Backend systems should not be called in the case that the field it is required for is skipped.

### Current behavior
Backends are called even if the field is skipped

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     },
     "coverageThreshold": {
       "global": {
-        "branches": 91,
+        "branches": 90,
         "functions": 96,
         "lines": 96,
         "statements": 96
--- a/src/__tests__/resolve-info.test.ts
+++ b/src/__tests__/resolve-info.test.ts
@@ -1310,14 +1310,7 @@ describe("resolver info", () => {
         resolvers: {
           Query: {
             node: (root, args, context, info) => {
-              // can be either Video or Media? or is it just have id?
-              // should we check there if this is Video or Media?
 
-              // i can't return interface.
-              // graphql returns only object types.
-              // if it doesn't sure what to return, it returns error.
-              // i can define `__resolveType()` in field so graphql
-              // would return this as default type for the field
               infNode = info;
               // eslint-disable-next-line no-prototype-builtins
               const lookaheadForUrl =
--- a/src/__tests__/resolve-info.test.ts
+++ b/src/__tests__/resolve-info.test.ts
@@ -1416,7 +1416,7 @@ describe("resolver info", () => {
             },
           }
         `);
-        console.log("starting query with the skip");
+
         doc = parse(`
             query($var: Boolean!) {
               node(id: "bla") {
--- a/src/__tests__/resolve-info.test.ts
+++ b/src/__tests__/resolve-info.test.ts
@@ -1265,6 +1265,7 @@ describe("resolver info", () => {
         });
         expect(idShouldInclude).toBe(false);
       });
+      
     });
   });
 });
--- a/src/__tests__/resolve-info.test.ts
+++ b/src/__tests__/resolve-info.test.ts
@@ -1357,12 +1357,14 @@ describe("resolver info", () => {
         return prepared.query(rootValue, contextValue, variableValues || {});
       }
 
+      beforeEach(() => {
+        expensiveFunction.mockClear();
+      })
+
       afterEach(() => {
         infNode = undefined;
         infElements = undefined;
         infMedia = undefined;
-
-        expensiveFunction.mockClear();
       });
 
       test("basic test", async () => {
--- a/src/__tests__/resolve-info.test.ts
+++ b/src/__tests__/resolve-info.test.ts
@@ -1234,7 +1234,7 @@ describe("resolver info", () => {
                   }
               `);
       });
-
+      // i guess i need to add few more tests for that with my own schema
       test("shouldInclude function is there and it works", async () => {
         const doc = parse(`
           query ($var: Boolean!) {
@@ -1266,5 +1266,75 @@ describe("resolver info", () => {
         expect(idShouldInclude).toBe(false);
       });
     });
+
+    describe.skip("lookahead resolution", () => {
+      let infNode: any;
+      let infElements: any;
+      let infMedia: any;
+      const schema = makeExecutableSchema(
+        {
+          typeDefs: `
+          type Query {
+            node(id: ID!): Node!
+          }
+
+          interface Node {
+            id: ID!
+          }
+
+          interface Media {
+            url: String!
+            tags: [Tag!]
+          }
+
+          type Image implements Node & Media {
+            id: ID!
+            url: String!
+            tags: [Tag!]
+            width: Int
+          }
+
+          type Video implements Node & Media {
+            id: ID!
+            url: String!
+            tags: [Tag!]
+
+          }
+
+        type Tag implements Node {
+          id: ID!
+          name: String!
+        }
+
+          `,
+          resolvers: {
+            Query: {
+              node: (root, args, context, info) => {
+                // can be either Video or Media? or is it just have id?
+                // should we check there if this is Video or Media?
+              }
+            },
+            Image: {
+              // eslint-disable-next-line @typescript-eslint/no-empty-function
+              url: (root, args, context, info) => {},
+              // eslint-disable-next-line @typescript-eslint/no-empty-function
+              tags: (root, args, context, info) => {},
+              // eslint-disable-next-line @typescript-eslint/no-empty-function
+              width: (root, args, context, info) => {}
+            },
+            Video: {
+              // eslint-disable-next-line @typescript-eslint/no-empty-function
+              url: (root, args, context, info) => {},
+              // eslint-disable-next-line @typescript-eslint/no-empty-function
+              tags: (root, args, context, info) => {}
+            },
+            Tag: {
+              // eslint-disable-next-line @typescript-eslint/no-empty-function
+              name: (root, args, context, info) => {}
+            },
+          }
+        }
+      )
+    })
   });
 });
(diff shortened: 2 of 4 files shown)


src/execution.ts
```diff
@@ -237,12 +240,14 @@ export function compileQuery<
     );
 
     let stringify: (v: any) => string;
-    if (options.customJSONSerializer) {
-      const jsonSchema = queryToJSONSchema(context);
-      stringify = fastJson(jsonSchema);
-    } else {
-      stringify = JSON.stringify;
-    }
+    // eslint-disable-next-line prefer-const
```
trkohler, line 243: i will remove it, this it for me for test locally

src/execution.ts
```diff
@@ -237,18 +237,21 @@ export function compileQuery<
     );
 
     let stringify: (v: any) => string;
-    if (options.customJSONSerializer) {
```
trkohler, line 240: this 100% can't be merged. It's just doesn't work for me locally.

src/variables.ts
```diff
@@ -57,7 +57,7 @@ export function compileVariableParsing(
   schema: GraphQLSchema,
   varDefNodes: ReadonlyArray<VariableDefinitionNode>
 ): (inputs: { [key: string]: any }) => CoercedVariableValues {
-  const errors = [];
```
trkohler, line 60: those arrays without the types are erroring on compilation.

it's also not so easy to give them a type, because they can have literally almost anything inside. they are not strongly typed arrays.

package.json
```diff
@@ -87,7 +87,8 @@
     "json-schema": "^0.4.0",
     "lodash.memoize": "^4.1.2",
     "lodash.merge": "4.6.2",
-    "lodash.mergewith": "4.6.2"
+    "lodash.mergewith": "4.6.2",
+    "perron": "^0.11.4"
```
trkohler, line 91: need to be changed to dev dependency in case someone would want to launch my script locally

trkohler:
it's possible for us in runtime to decide on lookahead

```
const lookaheadForName = info.fieldExpansion.Fruit.name.__shouldInclude({
        variables: info.variableValues
      });

console.log(`should we do lookahead for name? ${lookaheadForName}`)
```

src/resolve-info.ts
```diff
@@ -57,9 +67,13 @@ export interface TypeExpansion {
...
 }
 
-function createLeafField<T extends object>(props: T): T & LeafField {
+function createLeafField<T extends object>(
+  props: T,
+  shouldInclude: (__context: ShouldIncludeContext) => boolean
+): T & LeafField {
   return {
     [LeafFieldSymbol]: true,
+    __shouldInclude: shouldInclude,
```
boopathi, line 76: this function can be moved to the object's prototype instead of property, so it does not appear when iterating over the keys of the object.

src/resolve-info.ts
```diff
@@ -182,18 +195,30 @@ const memoizedExpandFieldNode = memoize4(expandFieldNode);
 function expandFieldNode(
   schema: GraphQLSchema,
   fragments: FragmentsType,
-  node: FieldNode,
+  node: JitFieldNode,
   fieldType: GraphQLOutputType
 ): FieldExpansion | LeafField {
+  const shouldInclude = (__context: ShouldIncludeContext): boolean => {
```
boopathi, line 201: We should not expect the caller to pass the context. Maybe only accepting variables is good.

trkohler: I've changes the type and renamed arg to `variables`. Should be more convenient

src/resolve-info.ts
trkohler: moved to prototypes and updated snapshots, might be better

trkohler:
current problem

this test with snapshot is not passing `src/__tests__/resolve-info.test.ts:902`

there is query 

```gql
query {
          node(id: "root") {
            id
            ... on Image {
              width
            }
            ... on DocumentElement {
              __typename
              ... on Image {
                url
                tags {
                  name
                }
              }
              ... on Media {
                tags {
                  id
                }
              }
            }
          }
        }
```

and I receive that my result contains property `name` but apparently it should not

```js
"tags": Object {
            "Tag": Object {
              "id": Object {
                Symbol(LeafFieldSymbol): true,
              },
    -         "name": Object {
    -           Symbol(LeafFieldSymbol): true,
    -         },
  ```
but we query property `name` on `tags`, aren't we? why we shouldn't have this property in snapshot?

**Comment by oporkka** [comment]
> current problem
> 
> this test with snapshot is not passing `src/__tests__/resolve-info.test.ts:902`
> 
> there is query
> 
> ```graphql
> query {
>           node(id: "root") {
>             id
>             ... on Image {
>               width
>             }
>             ... on DocumentElement {
>               __typename
>               ... on Image {
>                 url
>                 tags {
>                   name
>                 }
>               }
>               ... on Media {
>                 tags {
>                   id
>                 }
>               }
>             }
>           }
>         }
> ```
> 
> and I receive that my result contains property `name` but apparently it should not
> 
> ```js
> "tags": Object {
>             "Tag": Object {
>               "id": Object {
>                 Symbol(LeafFieldSymbol): true,
>               },
>     -         "name": Object {
>     -           Symbol(LeafFieldSymbol): true,
>     -         },
> ```
> 
> but we query property `name` on `tags`, aren't we? why we shouldn't have this property in snapshot?

I re-read the comment now, and realised it is other way around. The snapshot is expecting `name` to be there, but the result returned by `infNode.fieldExpansion` does not have it. Would that mean that the PR caused some sort of regression which the tests are now catching?

**Comment by trkohler** [comment]
yes, that's why it certainly must not be merged and I need to find a place where it's happening.

File: src/resolve-info.ts [hunk 7]
```diff
@@ -196,18 +231,31 @@ const memoizedExpandFieldNode = MEMOIZATION
...
   const typ = memoizedResolveEndType(fieldType) as GraphQLCompositeType;
   const possibleTypes = memoizedGetPossibleTypes(schema, typ);
 
-  const fieldExpansion: FieldExpansion = {};
+  const fieldExpansion: FieldExpansion = Object.create(
+    {
+      __shouldInclude: shouldInclude
+    },
+    {}
+  );
```
ponimas, line 258: I have 0 ideas how, but these lines break objects merging and lead to the failing test in this PR.

ponimas, line 258: I have 0 ideas how, but these lines break objects merging and lead to the failing test in this PR.

oporkka: The object merging breaks because we set here a prototype for the object, but lodash `mergeWith` is expecting objects to be plain objects to function, and it does not know how to merge objects that have non `null` prototypes.

See https://lodash.com/docs/4.17.15#merge

> Array and plain object properties are merged recursively. Other objects and value types are overridden by assignment.

Basically because we are not having plain objects anymore, in this code https://github.com/lodash/lodash/blob/2da024c3b4f9947a48517639de7560457cd4ec6c/.internal/baseMergeDeep.js#L69-L89 we skip this merge part, and just override the value, as stated above

oporkka: Our `customizer` for `mergeWith` is only customizing the leaf nodes, and returns `undefined` for other nodes, i.e. then normal lodash merge is used.

One way to deal with this would be to extend the customizer function to also merge our new objects that have prototype https://github.com/zalando-incubator/graphql-jit/blob/9de816e29801eb692f51993ecdfeb171a6d44ae8/src/resolve-info.ts#L378-L393

src/resolve-info.ts
```diff
@@ -59,11 +74,27 @@ export interface TypeExpansion {
...
 }
 
-function createLeafField<T extends object>(props: T): T & LeafField {
-  return {
-    [LeafFieldSymbol]: true,
-    ...props
-  };
+function createLeafField<T extends object>(
+  props: T,
+  shouldInclude: (variables: ShouldIncludeVariables) => boolean
```
boopathi, line 79: The typename "ShouldIncludeVariables" should be `ShouldIncludeParams`. Otherwise it would read `variables.variables[variableName]`.

src/__tests__/resolve-info.test.ts
```diff
@@ -1234,6 +1234,37 @@ describe("resolver info", () => {
...
+            }
+        `);
+        const rootValue = { root: "val" };
+
+        await executeQuery(schema, doc, rootValue, null, {
+          var: true
+        });
+        const validationErrors = validate(schema, doc);
+        if (validationErrors.length > 0) {
+          console.error(validationErrors);
```
boopathi, line 1258: instead of logging, add an expect statement to have validation errors are empty.

src/resolve-info.ts
```diff
@@ -37,19 +39,32 @@ export interface ResolveInfoEnricherInput {
...
+
+export type FieldExpansion = ShouldIncludeExtension & {
   // The possible return types that the field can return
   // It includes all the types in the Schema that intersect with the actual return type
   // eslint-disable-next-line no-use-before-define
   [returnType: string]: TypeExpansion;
-}
+};
+
+type RootFieldExpansion = {
```
boopathi, line 60: Why do we need this separate type?

trkohler: initially it was because root doesn't have ShouldIncludeExtension since root is always included.

trkohler:
closing because I would direct changes to internal version of the library since they are not needed in public version anyway

trkohler:
current todo

- [ ] exposing function in new attribute would be reflected in snapshots. need to figure out this or introduce those changes only in Zalando private fork of library (???)
- [ ] tests are very simple now... can be much better

src/resolve-info.ts
trkohler: renamed.

PR #180, opened 2022-09-18
fix: field expansion - fill only relevant fields while traversing abstract types

### LINKED ISSUE #112: Field Expansion for Union and Interfaces misbehaving
**Description**
When Union or  interfaces are sharing the same  types `too` as in the bellow schema, `too` is expanded for all types(Foo and Bar),  We expect shared fields are expand only for fields used in query

Given Schema
``` GQL

        type Query {
          uBaz: Baz
        }
        union Baz = Foo | Bar
        type Foo {
          foo: String
          too: String
        }
        type Bar {
          bar: Int
          too: String
        }
      `
```
and Query
```GQL
 `
            query {
              uBaz {
                ... on Foo {
                  foo
                  too
                }
                ... on Bar {
                  bar
                }
              }
            }
          `
```

Expected `info.fieldExpansion`  is

```
                  Object {
                    "Bar": Object {
                      "bar": Object {
                        Symbol(LeafFieldSymbol): true,
                      },
                    },
                    "Foo": Object {
                      "foo": Object {
                        Symbol(LeafFieldSymbol): true,
                      },
                      "too": Object {
                        Symbol(LeafFieldSymbol): true,
                      },
                    },
                  }
```

But Actual  `info.fieldExpansion`  is
```
                 Object {
                    "Bar": Object {
                      "bar": Object {
                        Symbol(LeafFieldSymbol): true,
                      },
                      "too": Object {     <--------- `too` is not expected as its not in the query   
                        Symbol(LeafFieldSymbol): true,
                      },
                    },
                    "Foo": Object {
                      "foo": Object {
                        Symbol(LeafFieldSymbol): true,
                      },
                      "too": Object {
                        Symbol(LeafFieldSymbol): true,
                      },
                    },
                  }
           
```

boopathi:
fix #112

### DIFF
--- a/src/resolve-info.ts
+++ b/src/resolve-info.ts
@@ -174,7 +174,7 @@ type FragmentsType = GraphQLResolveInfo["fragments"];
 type GraphQLNamedOutputType = GraphQLNamedType & GraphQLOutputType;
 type GraphQLObjectLike = GraphQLInterfaceType | GraphQLObjectType;
 
-const MEMOIZATION = false;
+const MEMOIZATION = true;
 
 const memoizedGetReturnType = MEMOIZATION
   ? memoize2(getReturnType)
--- a/src/resolve-info.ts
+++ b/src/resolve-info.ts
@@ -10,11 +10,13 @@ import {
   GraphQLOutputType,
   GraphQLResolveInfo,
   GraphQLSchema,
+  isAbstractType,
   isCompositeType,
   isListType,
   isNonNullType,
   isObjectType,
   isUnionType,
+  Kind,
   SelectionSetNode
 } from "graphql";
 import memoize from "lodash.memoize";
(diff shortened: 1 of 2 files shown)


boopathi:
👍

src/resolve-info.ts
```diff
@@ -172,12 +174,24 @@ type FragmentsType = GraphQLResolveInfo["fragments"];
 type GraphQLNamedOutputType = GraphQLNamedType & GraphQLOutputType;
 type GraphQLObjectLike = GraphQLInterfaceType | GraphQLObjectType;
 
-const memoizedGetReturnType = memoize2(getReturnType);
-const memoizedHasField = memoize2(hasField);
-const memoizedResolveEndType = memoize(resolveEndType);
-const memoizedGetPossibleTypes = memoize2(getPossibleTypes);
-const memoizedExpandFieldNodeType = memoize4(expandFieldNodeType);
-const memoizedExpandFieldNode = memoize4(expandFieldNode);
+const MEMOIZATION = true;
```
ponimas, line 177: Could you explain why we need that constant?

boopathi: Only for debugging purposes when changing / updating the implementation of field filtering.

topicus:
👍

Issue #184, opened 2022-10-26
GraphQL 17 support?


igrlk:
Hey everyone ✋. I'm currently working on adding `@defer` directive support to Mercurius GraphQL server [here](https://github.com/mercurius-js/mercurius/pull/898).
And it looks like some of the tests in Mercurius where it interacts with graphql-jit, and most of the tests in this repo are failing against the new major version of GraphQL (version 17, which is in alpha yet).

Are there any plans on adding support for `graphql@17`? Also, how open are you to the PRs from the community to add support for `graphql@17`, in case someone will want to contribute?

Looking forward to hearing from you. Thanks! 👍

boopathi:
https://github.com/zalando-incubator/graphql-jit/pull/185 is merged.

PR #185, opened 2022-11-14
Graphql 17 support

igrlk:
👋 Hi. I didn't get a response for [this question about graphql@17 support](https://github.com/zalando-incubator/graphql-jit/issues/184) and decided to go ahead and make `graphql-jit` work with the latest `graphql-js`

This PR stops using things that were deprecated in `graphql@17` and doesn't add any new functionality like `@defer` support - thankfully `graphql spec` says that [it's not required to support @stream/@defer](https://github.com/graphql/graphql-spec/pull/742/files#diff-30a69c5a5eded8e1aea52e53dad1181e6ec8f549ca2c50570b035153e2de1c43R1902) directives to be fully spec compatible. 

I think we don't need to merge it into main branch until `graphql@17` is released

Happy to hear any feedback from the contributors on this. Cheers! 🎩


oporkka:
Thanks for your contribution @igrlk. We added already test runs for the alpha version of 17. It seems that the tests that run also with GraphQL 15 and GraphQL 16 need fixing. Do you want to look into it to add support for older versions? We have version-specific code in style https://github.com/zalando-incubator/graphql-jit/blob/3cbb8b1586339ca028bb6b7b68f4b1294e55e715/src/__tests__/json.test.ts#L133-L139

igrlk:
@oporkka thanks for checking this PR! Yep I'm happy to fix the tests for other versions - would you mind allowing me to run the github checks? It will make the process faster :)

igrlk:
Hey @oporkka I was able to get the tests green on node 14, 16, 18 in combo with graphql 15, 16, 17 locally. Pushed all the changes to the branch

oporkka:
Tests allowed, but fail on coverage. I am not sure why they start failing after the change, is it actually the support for version info that causes the branch coverage to degrade? In any case, propose to lower the target for now to 91% to make all the tests run

igrlk:
Decreased the coverage threshold for branches to 91.

Still can't run the tests btw, looks like I need to become a contributor in the repo first to be able to run them

igrlk:
@oporkka thanks for running the tests. Since I've added the additional `versionInfo.major` checks and the package still works with the old versions of graphql, do you think it's something we can merge now? Happy to hear your thoughts on that

src/ast.ts
```diff
@@ -541,10 +543,25 @@ export function resolveFieldDef(
...
     const fieldName = fieldNode.name.value;
-    return getFieldDef(compilationContext.schema, parentType, fieldName as any);
+    return (execute as any).getFieldDef(
+      compilationContext.schema,
+      parentType,
+      fieldName as any
+    );
   }
 
-  return getFieldDef(compilationContext.schema, parentType, fieldNode as any);
```
boopathi, line 547: Let's add a shim here (https://github.com/zalando-incubator/graphql-jit/blob/main/src/compat.ts) in this repo to work around this change.

src/format-error.ts
```diff
@@ -0,0 +1,11 @@
+import { GraphQLError, versionInfo } from "graphql";
+import * as utilities from "graphql/error";
+import { GraphQLFormattedError } from "graphql/error";
+
+export const formatError = (error: GraphQLError): GraphQLFormattedError => {
```
boopathi, line 5: nit: please refactor to function declaration instead of using arrow function expression.

src/get-graphql-error-options.ts
```diff
@@ -0,0 +1,13 @@
+import { Maybe } from "graphql/jsutils/Maybe";
+import { ASTNode } from "graphql/language/ast";
+import { GraphQLError, versionInfo } from "graphql";
+
+export const getGraphQLErrorOptions = (
```
boopathi, line 5: move this to the `src/compat.ts` file

src/compat.ts
```diff
@@ -14,14 +17,20 @@ import * as GraphQL from "graphql";
...
-  if (GraphQL.getOperationRootType) {
-    return GraphQL.getOperationRootType(schema, operation);
-  } else {
-    // the use of any is to support graphql v15 types which will not use this codepath
-    return (schema as any).getRootType(operation.operation)!;
+
+export const getOperationRootType = (
+  schema: GraphQLSchema,
+  operation: OperationDefinitionNode
+): GraphQLObjectType => {
```
boopathi, line 24: nit: please refactor this arrow function expression to function declaration.

package.json
```diff
@@ -39,18 +39,18 @@
...
+        "branches": 91,
         "functions": 96,
         "lines": 96,
         "statements": 96
       }
     }
   },
   "peerDependencies": {
-    "graphql": ">=15"
+    "graphql": ">=17"
```
oporkka, line 50: This we should keep as `>=15` for now, as we still support 15

package.json
```diff
@@ -72,7 +72,7 @@
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-promise": "^5.1.1",
     "eslint-plugin-standard": "^5.0.0",
-    "graphql": "^16.0.0",
+    "graphql": "^17.0.0-alpha.2",
```
oporkka, line 75: This could be also changed back to 16 before merging, as 17 is nevertheless still alpha. We will keep support in test runs, though.

src/ast.ts
```diff
@@ -526,26 +525,7 @@ function getFieldEntryKey(node: FieldNode): string {
...
-  const fieldNode = fieldNodes[0];
-
-  if (versionInfo.major < 16) {
-    const fieldName = fieldNode.name.value;
-    return getFieldDef(compilationContext.schema, parentType, fieldName as any);
-  }
-
-  return getFieldDef(compilationContext.schema, parentType, fieldNode as any);
-}
+export { resolveFieldDef };
```
igrlk, line 528: Decided to leave this file as an entry point for `resolveFieldDef`. @boopathi wdyt?

igrlk:
Thank you for your comments, @boopathi @oporkka. All the things you mentioned should be fixed now

igrlk:
Fixed the tests

.idea/modules.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
oporkka, line 1: Let's remove this file. I propose to have `.idea/` in your global .gitignore file.

oporkka:
:+1:

igrlk:
Hi @boopathi @oporkka. Is there something that's blocking us from merging this PR? 😄

boopathi (approved):

boopathi:
👍

oporkka:
> Hi @boopathi @oporkka. Is there something that's blocking us from merging this PR? 😄

Sorry for the delay, we just forgot to follow up on this as other things popped up.

PR #188, opened 2023-01-10
Drop build for Node 12, add 18

oporkka:
Node version 12 is end of lifetime, so drop that from build jobs.

The current LTS version is 18, so add that to build jobs.


ruiaraujo:
👍

boopathi:
👍

oporkka:
:+1:

PR #193, opened 2023-01-30
fix: validation for skip/include with default values

### LINKED ISSUE #190: graphql-jit rejecting nullable parameter with default values in @include directive
Copied from https://github.com/n1ru4l/envelop/issues/1615

### Issue workflow progress

_Progress of the issue based on the [Contributor Workflow](https://github.com/the-guild-org/Stack/blob/master/CONTRIBUTING.md#a-typical-contributor-workflow)_

- [ ] 1. The issue provides a [minimal reproduction](https://en.wikipedia.org/wiki/Minimal_reproducible_example) available on [Stackblitz](https://stackblitz.com/fork/node).
  - _Please install the latest `@envelop/*` packages that you are using._
  - _Please make sure the reproduction is as small as possible._
- [ ] 2. A failing test has been provided
- [ ] 3. A local solution has been provided
- [ ] 4. A pull request is pending review

---

**Describe the bug**

Nullable arguments to a query with a default values are considered non-nullable. This, however, is not the case for variables used in directives (specifically tested with `@include(if: )`:
```graphql
query myQuery($inc: Boolean = true) {
  myField(id: 5) @include(if: $inc)
}
```

This yields the following error:
```
Variable 'inc' of type 'Boolean' used in position expecting type 'Boolean!'
```

**To Reproduce**

See above.

**Expected behavior**

The error should not occur, and the directive should be evaluated according to the variable's default value.

**Environment:**

```
    "@envelop/core": "^3.0.4",
    "@envelop/graphql-jit": "^5.0.4",
    "@envelop/newrelic": "5.0.4",
    "@envelop/sentry": "^4.0.4",
```

- OS: `Ubuntu 22.04.1 LTS (WSL 2)`
- NodeJS: `v18.12.1`

**Additional context**

n/a

boopathi:
fix #190

Corresponding spec text: https://spec.graphql.org/June2018/#sec-All-Variable-Usages-are-Allowed

When validation the variable definition for the argument `if` for the directives skip/include, the default values in the variable definition were not considered. This PR fixes the issue by correctly using the default values in the variable definition.

### DIFF
--- a/src/__tests__/directives.test.ts
+++ b/src/__tests__/directives.test.ts
@@ -1383,4 +1383,54 @@ describe("Execute: handles directives", () => {
       expect(mockResolver).toHaveBeenCalled();
     });
   });
+
+  describe("default values in include / skip directive's if", () => {
+    test("default value true in include", () => {
+      const result = executeTestQuery(`
+        query myQuery($inc: Boolean = true) {
+          a @include(if: $inc)
+        }
+      `);
+
+      expect(result).toEqual({
+        data: { a: "a" }
+      });
+    });
+
+    test("default value false in include", () => {
+      const result = executeTestQuery(`
+        query myQuery($inc: Boolean = false) {
+          a @include(if: $inc)
+        }
+      `);
+
+      expect(result).toEqual({
+        data: {}
+      });
+    });
+
+    test("default value true in skip", () => {
+      const result = executeTestQuery(`
+        query myQuery($skip: Boolean = true) {
+          a @skip(if: $skip)
+        }
+      `);
+
+      expect(result).toEqual({
+        data: {}
+      });
+    });
+
+    test("default value false in skip", () => {
+      const result = executeTestQuery(`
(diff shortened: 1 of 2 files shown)


trkohler:
👍

src/__tests__/directives.test.ts
```diff
@@ -1383,4 +1383,54 @@ describe("Execute: handles directives", () => {
...
+          a @skip(if: $skip)
+        }
+      `);
+
+      expect(result).toEqual({
+        data: {}
+      });
+    });
+
+    test("default value false in skip", () => {
```
trkohler, line 1424: maybe also add test when there is a variable passed to the query and query has also default?

trkohler: like 

```
query bla(a: Boolean = True) { ... }

variables(a: False)
```

boopathi: those tests are already present in variables.test.ts

oporkka:
:+1:

PR #197, opened 2023-03-13
fix: skip include logic handling multiple paths to same node

### LINKED ISSUE #166: misbehaving spreading fragments
Given Schema:
```
type Query {
  detailContent: [Content!]
}


type Post implements Content {
  id: ID!
  title: String!
  type: String!
  related: [Content]
}

type Article implements Content {
  id: ID!
  title: String!
  type: String!
}

interface Content {
  id: ID!
  title: String!
  type: String!
}
```
and query: 

```
query TEST(
    $includeOnly: Boolean!
){
  detailContent{
    ...articleFragment
    ...on Post {
      ...postFragment 
      related{
          id
  	  ...articleFragment @include(if:$includeOnly)
    	}
    }
  }
}

fragment postFragment on Post {
  id
  title
}

fragment articleFragment on Article {
title
  type
}
```

Expect:
```
{
  "data": {
    "detailContent": [
      {
        "id": "post:1",
        "title": "Introduction to GraphQL!",
        "related": [
          {
            "id": "article:1",
          }
        ]
      },
      {
        "id": "post:2",
        "title": "GraphQL-Jit a fast engine for GraphQL",
        "related": [
          {
            "id": "article:2",
          }
        ]
      },
      {
        "title": "article Introduction to GraphQL!",
        "type": "article"
      },
      {
        "title": "article GraphQL-Jit a fast engine for GraphQL",
        "type": "article"
      }
    ]
  }
}
```
But receive:
```
{
  "data": {
    "detailContent": [
      {
        "id": "post:1",
        "title": "Introduction to GraphQL!",
        "related": [
          {}
        ]
      },
      {
        "id": "post:2",
        "title": "GraphQL-Jit a fast engine for GraphQL",
        "related": [
          {}
        ]
      },
      {},
      {}
    ]
  }
}
```

boopathi:
- fix #166 

When there are multiple paths to the same field Node (fragments), there can be multiple skip/include requirements for the same node. This breaks our assumption in the existing implementation where we compute the skip/include for a particular field node and use the same for all paths.

For example, 

```gql
query {
  paths {
    one {
      ...frag @skip(if: true)
    }
    two {
      ...frag @skip(if: false)
    }
  }
}
fragment frag on Item {
  target
}
```

Here, in this query, the field `target` has two paths - `paths.one.target`, and `paths.two.target`. But, in the current implementation we compute `__internalShouldInclude` for a particular field node - so this gets reused for all paths to the field Node - so is the reason for the bug.

This PR fixes it by remodelling the internal representation. This PR uses a path specific skip/include logic and while generating code, it uses the path specific logic. We represent this as an object - `__internalShouldIncludePaths[path.in.dot.notation] = SkipIncludeCode` for that path.

TODO:

- [ ] Move tests to directives.test.ts
- [ ] Remove unused code - the old `__internalShouldInclude`
- [ ] Remove comments

### DIFF
--- a/src/ast.ts
+++ b/src/ast.ts
@@ -33,9 +33,23 @@ import { isAbstractType } from "graphql/type";
 import { CompilationContext, GLOBAL_VARIABLES_NAME } from "./execution";
 import createInspect from "./inspect";
 import { getGraphQLErrorOptions, resolveFieldDef } from "./compat";
+import { memoize4 } from "./memoize";
 
 export interface JitFieldNode extends FieldNode {
+  /**
+   * @deprecated Use __internalShouldIncludePath instead
+   * @see __internalShouldIncludePath
+   */
   __internalShouldInclude?: string;
+
+  // The shouldInclude logic is specific to the current path
+  // This is because the same fieldNode can be reached from different paths
+  // - for example, the same fragment used in two different places
+  __internalShouldIncludePath?: {
+    // Key is the stringified ObjectPath,
+    // Value is the shouldInclude logic for that path
+    [path: string]: string;
+  };
 }
 
 export interface FieldsAndNodes {
@@ -57,14 +71,17 @@ export function collectFields(
   runtimeType: GraphQLObjectType,
   selectionSet: SelectionSetNode,
   fields: FieldsAndNodes,
-  visitedFragmentNames: { [key: string]: boolean }
+  visitedFragmentNames: { [key: string]: boolean },
+  previousPath?: ObjectPath
 ): FieldsAndNodes {
   return collectFieldsImpl(
     compilationContext,
     runtimeType,
     selectionSet,
     fields,
-    visitedFragmentNames
+    visitedFragmentNames,
+    undefined,
+    serializeObjectPathForSkipInclude(previousPath)
   );
 }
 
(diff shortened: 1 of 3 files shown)


.eslintrc.json
```diff
@@ -18,7 +18,8 @@
     "@typescript-eslint/interface-name-prefix": "off",
     "comma-dangle": "off",
     "sort-keys": "off",
-    "max-classes-per-file": "off"
+    "max-classes-per-file": "off",
+    "no-use-before-define": "off"
```
oporkka, line 22: Why do we want to set this one off?

boopathi: This one makes no sense for most cases - e.g. type definitions - it is always okay to use a type definition before declaration, and also function declaration statements. The TDD (Temporal Dead Zone) detection is much better in TS compiler for the relevant ones. So, this rule in practice was just annoyance rather than preventing any potential bugs.

ayruslore:
👍

src/ast.ts
```diff
@@ -57,14 +71,17 @@ export function collectFields(
   runtimeType: GraphQLObjectType,
   selectionSet: SelectionSetNode,
   fields: FieldsAndNodes,
-  visitedFragmentNames: { [key: string]: boolean }
+  visitedFragmentNames: { [key: string]: boolean },
+  previousPath?: ObjectPath
```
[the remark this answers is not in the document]
boopathi: Done!

boopathi:
👍

oporkka:
:+1:

PR #200, opened 2023-04-17
fix: use option to compute path based skip/include

boopathi:
The previous fixes and additions to skip/include for fragment spreads #197 causes high event loop lag during compilation. This change is to compute the path based skip include only if the compiler option is set. Previously, we used the compiler option to only use it in the generated output - now it is also used for computation.


ayruslore:
👍

boopathi:
👍

oporkka:
:+1:

PR #202, opened 2023-05-04
Build on Node 20, drop 14

oporkka:
Node 14 is EOL, and Node 20 is now the current version.

Run tests on Node 20 instead of Node 14

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


oporkka:
> We can also update the TypeScript target version too along with this - https://github.com/zalando-incubator/graphql-jit/blob/main/tsconfig.json#L4

We can do one step at time

boopathi (approved):

oporkka:
:+1:

PR #204, opened 2023-05-10
chore: upgrade dev deps

trim21:
some eslint plugins are too old so it cause peerDeps mismatch.

(why there is only lint-staged without husky, should we add husky too?)


trim21:
I find that I can't upgrade stryker to latest version, I'll just close this.

PR #205, opened 2023-06-14
Bump es target version to es2021

oporkka:
Now that minimum supported Node version is 16, we can target es2021 (ref. https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping)

Note: This PR fails on build for Node 18 and Node 20 because "Support for the experimental syntax 'optionalChaining' isn't currently enabled". The problem will get fixed by bumping jest to the latest version, which in turn  includes `@babel/plugin-syntax-optional-chaining`


oporkka:
:+1:

boopathi:
👍

PR #206, opened 2023-06-14
Update jest from 27 to latest 29

oporkka:
Updated jest, jest types and ts-jest to the latest versions.

The upgrade to v29 includes a breaking change on snapshots so they are updated here as well (ref. https://jestjs.io/blog/2022/08/25/jest-29)


boopathi:
👍

oporkka:
:+1:

Issue #208, opened 2023-06-21
TypeError: Cannot set properties of null (setting 'fieldName')


bbnjmn:
Hi everyone and thanks for open-sourcing this project 🙂

## Issue

We’ve been facing flaky issues on some queries that sometimes leads to a `TypeError: Cannot set properties of null (setting 'fieldName')`. After some investigations it seems this depends on the order in which fields are resolved, that’s why we don't get the issue all the time.

## How to reproduce ?

Here is an example of a query that could sometimes trigger the issue :

```graphql
{
  rootField {
    nestedField {
      nonNullableFieldThatMightResolveAfterTheOthers
    }
    nonNullableField
  }
}
```

_For the examples we'll assume that `nonNullableFieldThatMightResolveAfterTheOthers` and `nonNullableField ` are both returning `null`._

**If fields resolution happens in that order :**
1. nonNullableFieldThatMightResolveAfterTheOthers
2. nonNullableField

We get the expected response
```json
{
  "data": {
    "rootField": null
  }
}
```

**If fields resolution happens in that order:**
1. nonNullableField
2. nonNullableFieldThatMightResolveAfterTheOthers

It triggers a `TypeError: Cannot set properties of null (setting 'nestedField')`.

PR #209, opened 2023-06-21
fix: non-null field returning null sometimes triggers TypeError

bbnjmn:
* fix https://github.com/zalando-incubator/graphql-jit/issues/208

## Bug description

Under some fields resolution timing conditions, non null fields that returns nulls may sometime trigger a `TypeError` and make a whole query fail.

This happens in[`removeBranch` function](https://github.com/zalando-incubator/graphql-jit/blob/f20e6da017a3819029c24fe3349cec3ff8cdc255/src/non-null.ts#LL84C10-L84C10) . If we are trying to remove a branch of `tree`  that has already been removed by a non nullable field (a sibling of any ancestor) because it has returned `null`.

## Proposed solution
We can protect us from `TypeError` by ignoring branches that are no longer part of the tree when calling `removeBranch`. Error will be preserved and returned back to the client to indicate that a non null field returned a null value (= same behavior as currently).

Note : I’ve split my PR into 2 commits so the first one includes only tests that are failing currently, and the second one actually contains the fix.

### DIFF
--- a/src/__tests__/nonnull.test.ts
+++ b/src/__tests__/nonnull.test.ts
@@ -37,12 +37,12 @@ const throwingData = {
     });
   },
   latePromise() {
-    return new Promise((_, reject) =>
+    return new Promise((resolve, reject) =>
       setTimeout(() => reject(latePromiseError), 0)
     );
   },
   latePromiseNonNull() {
-    return new Promise((_, reject) =>
+    return new Promise((resolve, reject) =>
       setTimeout(() => reject(latePromiseNonNullError), 0)
     );
   },


boopathi:
Thank you for your contributions with this PR. I will conduct local testing to understand the bug and its potential effects more thoroughly before continuing with the PR review.

boopathi (approved):

boopathi:
👍

bbnjmn:
Hello @boopathi,

How can we make this PR progress as it seems it requires an additional approval ?

Thanks in advance 🙂

bbnjmn:
@ruiaraujo Can you check this out? Thanks 🙇

oporkka:
:+1:

oporkka:
@bbnjmn thanks for the contribution and sorry to let you wait. I triggered already a new release

PR #210, opened 2023-06-29
Move object initalisation behind a feature toggle

oporkka:
Moved the initialisation of __internalShouldIncludePath under field node to happen behind the configuration option
compilationContext.options.useExperimentalPathBasedSkipInclude.

The context for the change is that updating from 0.8.0 to 0.8.2 has shown increased Garbage Collector rates in use cases with big queries, and this change could be used to verify the hypothesis whether the initialisation of __internalShouldIncludePath is a cause for the GC runs.

### DIFF
--- a/src/ast.ts
+++ b/src/ast.ts
@@ -123,9 +123,6 @@ function collectFieldsImpl(
           selection
         );
 
-        if (!fieldNode.__internalShouldIncludePath)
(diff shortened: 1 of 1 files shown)


oporkka:
:+1:

boopathi:
👍

PR #213, opened 2023-06-29
Use Node16 actions in publish.yml

oporkka:
The commands `actions/checkout@v2` and `actions/setup-node@v2` are deprecated. We should use the next version of those actions 

- Announcement https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/
- How to use versions https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-versioned-actions

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -6,9 +6,9 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


boopathi:
👍

oporkka:
:+1:

PR #214, opened 2023-07-03
remove memoization to reduce memory pressure

nabeelbukhari:
We observed recently that this function - 

1. does not hit the cache . 100% of the time for 4 parameters, the cache miss happens
2. It causes a lot of map objects - as a result of the cache miss, the nested map would create for more variance - and causes application to be slow.


boopathi:
👍

boopathi (approved):

nabeelbukhari:
👍

PR #222, opened 2023-09-14
docs: add more info about the differences to GraphQL-JS

### LINKED ISSUE #221: README.md not clear on consequences of limitations
At https://github.com/zalando-incubator/graphql-jit/#differences-to-graphql-js, the README.md mentions that

> The primary limitation is that all computed properties must have a resolver and only these can return a `Promise`.

However, it doesn't mention what the consequences are of failing to meet these criteria. How can they be violated? Will there be an error if they are violated, or will execution fall back to non-jitted? Could you please elaborate on the consequences of the limitations?

boopathi:
fix #221

### DIFF
--- a/GraphQL-JS.md
+++ b/GraphQL-JS.md
@@ -1,0 +1,45 @@
+## Differences to GraphQL-JS
+
+In order to achieve better performance, the `graphql-jit` compiler introduces some limitations.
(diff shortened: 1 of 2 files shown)


boopathi:
👍

oporkka:
:+1:

PR #226, opened 2023-09-14
feat: output both esm and cjs; upgrade dependencies

boopathi:
Part of overall maintenance of the project -

1. Output both CJS and ESM. Refer package.json.
2. Use faster `@swc/jest` instead of `ts-jest`
3. Use faster `@swc-node/register` instead of `ts-node` 
4. Upgrade all dependencies to their latest version

Nits:

1. Fix version of `wrap-ansi` to `7.x` under jest - as wrap-ansi published a version (8.x) that is ESM only and jest-cli using this running as CJS always errors out. I have given up trying to make esm work with jest and typescript.


oporkka:
:+1:

boopathi:
👍

PR #240, opened 2024-03-18
Bump build actions to v4

oporkka:
Bumped the build actions to v4 that uses Node20 by default, to avoid warnings of using outdated Node versions in actions

Updated the deprecated `set-output` to the new environment variables -based way to avoid warnings.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,9 +17,9 @@ jobs:
         graphql-version: [15, 16, 17.0.0-alpha.3]
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v4
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -17,20 +17,20 @@ jobs:
...
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node-version }}
 
       # https://github.com/actions/cache/blob/main/examples.md#node---yarn
       - name: Get yarn cache directory path
         id: yarn-cache-dir-path
-        run: echo "::set-output name=dir::$(yarn cache dir)"
+        run: echo "dir=$(yarn cache dir)" >> $GITHUB_OUTPUT
```
oporkka, line 29: set-output is deprecated. The new way is to use env vars https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/#examples

boopathi:
👍

oporkka:
:+1:

PR #248, opened 2024-06-25
Bump GraphQL 17 version in build

oporkka:
There is new alpha version 5 for GraphQL 17, so use it in builds to get results in the latest state of GraphQL 17.


oporkka:
:+1:

boopathi:
👍

PR #251, opened 2024-09-11
fix: export ResolveInfoEnricherInput

boopathi:
To use the `resolveInfoEnricher` functionality, it is required to export the type ResolveInfoEnricherInput, otherwise we find ourselves importing it from `graphql-jit/dist/resolve-info`.


oporkka (approved):

boopathi:
👍

PR #256, opened 2024-09-12
fix: create new canary version from published pkg

boopathi:
Fixes - 

1. Create a new canary version instead of recreating the same canary version - because we are skipping git tag and git commit for canary publishes
2. Fix the NPM_TOKEN variable name


boopathi:
👍

ayruslore (approved):

PR #257, opened 2024-09-13
fix: support esm; fix imports

boopathi:
Even with [new changes](https://github.com/zalando-incubator/graphql-jit/pull/252) it is buggy to use graphql-jit with esm. The problems are - file imports without extensions (e.g. "./execution"), internal package imports ("graphql/utils/Path"), and also directory imports ("graphql/language").

The other set of problems happen with esbuild (part of tsup) not being able to strip away type names from the build. For this it is required to mark the type imports with the type keyword in the import statements.

So, the PR fixes - 

1. use `.js` extension for imports that are not modules (relative imports and internal file / directory imports from packages)
2. add `type` modifier to all type only imports. This was done with the help of typescript 5's new feature [`verbatimModuleSyntax`](https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax). We are not preserving this TSConfig option because it requires the package to be of `type: module` which is not the case for us.
3. Because we add extension, we configure jest to remove the `.js` extension so that Jest is able to import the ts file directly.

As a follow up PR, we can validate in CI using tsc or other tool that we don't export types without the `type` modifier.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,9 @@
       "js",
       "json"
     ],
+    "moduleNameMapper": {
+      "(.+)\\.js": "$1"
+    },
     "testEnvironment": "node",
     "testRegex": "(/tests/.*|(\\.|/)test)\\.ts$",
     "transform": {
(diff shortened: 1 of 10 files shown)


ayruslore (approved):

boopathi:
👍

PR #258, opened 2024-09-13
chore: upgrade eslint to flat config

boopathi:
Upgrade to flat config for eslint and remove all legacy config and remove unnecessary eslint-disables


boopathi:
👍

PR #260, opened 2024-10-14
perf: micro optimizations

boopathi:
A few performance improvements to compilation based on micro-optimizations / benchmarks - 

1. string concatenation `body += ` seems to be faster than `[].join("\n")` that is used by generate-function.
2. Prefer looping over recursion as tail calls are not optimized in v8 as of this writing
3. Add memoization to simple functions - e.g. joinOriginPaths
4. Restructure compilations of skip/include to a `string[][]` and avoid the unnecessary string manipulation. Reduces complexity - simplifies filtering of unwanted `true`s and code generation.

Next set of optimizations will follow in separate PRs.

Observations with this - 

For a very big query (1100 lines formatted, 30+ fragments, 40+ variables, max depth - 8) that takes 1.7s to compile, this PR brings it down to around 1.4s - around 17% faster.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -26,7 +26,7 @@ export default tseslint.config(
       "no-sequences": "error",
       "no-template-curly-in-string": "error",
       "@typescript-eslint/no-empty-function": "error",
-      "@typescript-eslint/no-non-null-assertion": "error",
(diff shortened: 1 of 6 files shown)


ayruslore (approved):

boopathi:
👍

PR #261, opened 2024-10-14
perf: memoize skip include compilation

boopathi:
Work in progress.

Depends on #260. Right now it targets #260. After #260 is merged, I'll target branch main.

Memoize repeated computations of collect fields for the same sub-tree. To memoize, use a separate weakMemo based on WeakMap instead of using a memoize4 or memoize5 functions - this is because not all arguments of the function need to participate in the memo key. Only 2 out of 5 or 6 arguments actually contribute to the memo key.


oporkka:
Any edge cases we should check? Like against false caching?

PR #262, opened 2024-10-18
perf: alternate impl of getLocation

boopathi:
Improvement to getLocation to avoid the (1) regex of entire source, and (2) going over the entire source to figure out an information that already exists in node AST.

This improves the compilation time - compile time reduced by 60% - tested against a large query of 1100 lines, 30+ fragments, and 40+ variables.


ayruslore (approved):

boopathi:
👍

PR #270, opened 2026-01-20
fix: support latest version of graphql-17-alpha.9 release

boopathi:
Deprecate Node18, add Node24.

Handle breaking changes in graphql 17-alpha.9 release - 

1. `parseLiteral` → `coerceInputLiteral` - detect version and use the right function
2. `buildExecutionContext` removed. Replace it with `buildCompilationContext` in tests
3. Introspection Schema: `includeDeprecated` Now Required. Add it in the introspection query tests and in `getIntrospectionQuery({inputValueDeprecation: true})`

The build pipeline is split into two separate pipelines - 

1. `lint` - tsc, build, linter (matrix: none)
2. `test` - run tests (matrix: node 20,22,24; graphql 15,16,17-alpha)

This is because it is harder to achieve compatibility in the breaking type changes like GraphQLScalarType (generic vs no generic in older versions), ConstValueNode not available in older versions, etc. So we run build, and TSC only on current dev version in the repo (v16). But we run tests against all supported versions of graphql-js.

### DIFF
--- a/src/ast.ts
+++ b/src/ast.ts
@@ -1,6 +1,7 @@
 import {
   type ArgumentNode,
   type ASTNode,
+  type ConstValueNode,
   type DirectiveNode,
   type FieldNode,
   type FragmentDefinitionNode,
@@ -34,7 +35,11 @@ import {
 } from "graphql";
 import { type CompilationContext, GLOBAL_VARIABLES_NAME } from "./execution.js";
 import createInspect from "./inspect.js";
-import { getGraphQLErrorOptions, resolveFieldDef } from "./compat.js";
+import {
+  coerceInputLiteral,
+  getGraphQLErrorOptions,
+  resolveFieldDef
+} from "./compat.js";
 
 export interface JitFieldNode extends FieldNode {
   /**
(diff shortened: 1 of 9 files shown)


boopathi:
I have disabled the `build(<node-version>, <graphql-version>)` requirements for PR because of the changes to the build pipeline.

src/ast.ts
```diff
@@ -945,18 +949,14 @@ export function valueFromAST(
...
-      if (type.parseLiteral.length > 1) {
-        // eslint-disable-next-line
-        console.error(
-          "Scalar with variable inputs detected for parsing AST literals. This is not supported."
-        );
-      }
-      result = type.parseLiteral(valueNode, {});
+      // Use coerceInputLiteral if available (v17+), otherwise fall back to parseLiteral
+      // By this point, we've filtered out VARIABLE nodes, so valueNode is a ConstValueNode
+      result = coerceInputLiteral(type, valueNode as ConstValueNode);
```
oporkka, line 958: ~~Where is the fall back to parseLiteral happening?~~

I see it is inside `coerceInputLiteral` which is actually a wrapper. Could we rename it to something which seems obviously like a compatibility wrapper method, like `coerceInputOrParseLiteral` as long as we support both?

src/ast.ts
```diff
@@ -945,18 +949,14 @@ export function valueFromAST(
...
     let result: any;
     try {
-      if (type.parseLiteral.length > 1) {
-        // eslint-disable-next-line
-        console.error(
-          "Scalar with variable inputs detected for parsing AST literals. This is not supported."
-        );
-      }
-      result = type.parseLiteral(valueNode, {});
+      // Use coerceInputLiteral if available (v17+), otherwise fall back to parseLiteral
```
oporkka: Would this comment rather belong to the `coerceInputLiteral` function definition?

src/ast.ts
boopathi: In `compat.ts`

src/ast.ts
boopathi: Probably. This is a left over comment before I moved it to compat.ts. Will remove it.

src/ast.ts
boopathi: The usages are different - `coerceInputLiteral(type, args)` vs `type.coerceInputLiteral(args)`. It is similar to other compatibility methods - `getOperationRootType(schema, operation)` vs `schema.getOperationRootType(operation)`, and others in that file.

oporkka (approved):

boopathi:
👍

PR #275, opened 2026-03-23
fix: publish config for OIDC

boopathi:
- permissions inside the job
- set node version to 24 as npm min version to support OIDC is 11.5 and at least node 22 is required.
- condition that forks do not run publish
- Pr publish optional


dom-mel (approved):

boopathi:
👍

PR #277, opened 2026-05-11
Memory optimizations of compiled code

boopathi:
Profiling medium to large production compiled queries shows an average memory of 10MB per compiledQuery. This creates heap pressure and gc dominated event-loop. In this PR, we create a shared runtime module, extract common boilerplate code that exists in every query, and group common handlers into smaller versions to reduce the 10MB avg to 3.2MB for the same set of queries we started with.

#### Optimizations: 

1. Shared runtime extracts boilerplate code like promise handling, null checks, error checks, safeMap handling, resolver try catch into single module level object. We expose `ExecutionContext.rt` instead of inlining these patterns for every query, and these are shared across all compiled queries. 
2. Trivial skip/include ternaries are generated even for fields with no skip / include directives - `true ? value : undefined`. We eliminate this.
3. When the same field appears at the same query path multiple times within a single compilation, the error handling functions were simply duplicates. We deduplicate them in compilation context keyed on function body. This is most prominent in unions with N concrete types. 
4. Abstract types contain trivial cases that only include __typename field. Here, we generate a switch case per concrete type even when the bodies differ only in typename string literal. So, we add a precomputed finalType variable so that concrete types with no deferred resolvers can be grouped into a single case. A union type with a N (around 50 or 100) cases and handling only 2-3 cases in a query benefit a lot.

Measured on 120 queries, 

Avg per query: 
- Before: 10+ MB
- After: 3+ MB

Total:
- Before: 1200+ MB
- After: 380+ MB

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     }
   },
   "scripts": {
-    "benchmark": "NODE_ENV=production ./src/__benchmarks__/benchmarks.ts",
+    "benchmark": "NODE_ENV=production node --loader @swc-node/register/esm src/__benchmarks__/benchmarks.ts",
     "build": "yarn tsup-node",
     "check-exports": "attw --pack .",
     "check-tsc": "tsc --noEmit",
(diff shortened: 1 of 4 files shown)


dom-mel (approved):

boopathi:
👍

Issue #279, opened 2026-05-13
Regression in main branch


boopathi:
With the memory optimizations in https://github.com/zalando-incubator/graphql-jit/pull/277, we have a regression in the main branch for very large queries and they are timing out and are slow

boopathi:
#280 fixes this.

PR #280, opened 2026-05-18
fix: scope escape bug of finalType

boopathi:
fix for #279 

In the previous pr #277 , 

1. compileAbstractType compiles each type with `context.trivialTypeNameVar = "finalType"` so that trivial cases can share a single switch body.
2. but this context gets spread elsewhere in hoisted functions and generates `{ __typename: finalType, ...`  and results in a Reference Error during the onSuccess of resolver's promise. 
3. This happens after ++promiseCounter but before --promiseCounter. The counter was permanently stuck above 0. 

The fix is to set it to a local variable and immediately clear it in the context.

### DIFF
--- a/src/__tests__/abstract.test.ts
+++ b/src/__tests__/abstract.test.ts
@@ -732,6 +732,76 @@ describe("Abstract type trivial case deduplication", () => {
     );
   });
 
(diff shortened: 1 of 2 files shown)


boopathi:
👍

Abdelghafour01 (approved):

PR #282, opened 2026-06-17
support graphql-js v17

boopathi:
- drop Node 20, test against node 22,24,26
- Support graphql-js stable release v17
- set default dev version of graphql-js to v17


dom-mel (approved):

boopathi:
👍

PR #284, opened 2026-07-03
fix: tsconfig for node and jest globals

boopathi:
Editor shows red squiggly lines for NodeJS and Jest globals. Add these types to tsconfig.


boopathi:
👍

Abdelghafour01 (approved):

PR #290, opened 2026-09-15
feat: add virtual scripts for debugging compiled code

boopathi:
enables debugging of the generated / compiled js code for query and variables in vscode/chrome with launch.json or `node --inspect`

Examples - `simple-debug`, and `blog` use the debug.

Breaking changes:

- internal api `debug: boolean` now requires it to be an object `debug: { enabled: true, ...}`


boopathi:
Example of it working in Chrome with debugger paused before calling a resolver and showing the in-context args passed to the resolver. 

Screenshot 2026-09-15 at 18 16 53

axyz (approved):

boopathi:
👍

